PR #3, opened 2022-03-18
for nextest action, use version in URL

sunshowers:
Thanks for adding support for this

Nextest lets you specify the version in its URL. It also supports basic version ranges (e.g. 0.9, or later 1).

After this lands I'm going to point folks to this action as the recommended way to install nextest in GHA. Thanks again!


taiki-e (approved):
Thanks!

taiki-e:
Published in 1.2.2.

PR #8, opened 2022-05-31
Use `cargo-binstall` as fallback in `main.sh`

### LINKED ISSUE #7: Suggestion/Supporting cargo-binstall?
[`cargo-binstall`](https://github.com/ryankurte/cargo-binstall) seems like an awesome tool for CI to avoid building binaries.

It support manifest discovery via:
 - Fetch crate / manifest via [crates.io](crates.io)
 - Fetch build from the [quickinstall](https://github.com/alsuren/cargo-quickinstall) repository, which contains a lot of pre-compiled binary packages

NobodyXu:
Fixed #7 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,20 +26,20 @@ jobs:
           - ubuntu-20.04
         tool:
           # Note: Specifying the version of valgrind and wasm-pack is not supported.
-          # minifier is supported by cargo-binstall (through quickinstall)
-          - cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,valgrind,wasm-pack,wasmtime,mdbook,minifier
-          - cargo-hack@0.5.14,cargo-llvm-cov@0.4.5,cargo-minimal-versions@0.1.4,parse-changelog@0.4.8,cross@0.2.1,nextest@0.9.11,shellcheck@0.8.0,shfmt@3.5.0,wasmtime@0.37.0,mdbook@0.4.18,minifier@0.2.1
+          # cargo-watch is supported by cargo-binstall (through quickinstall)
+          - cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,valgrind,wasm-pack,wasmtime,mdbook,cargo-watch
+          - cargo-hack@0.5.14,cargo-llvm-cov@0.4.5,cargo-minimal-versions@0.1.4,parse-changelog@0.4.8,cross@0.2.1,nextest@0.9.11,shellcheck@0.8.0,shfmt@3.5.0,wasmtime@0.37.0,mdbook@0.4.18,cargo-watch@8.1.1
           # Nextest supports basic version ranges as well
           - nextest@0.9
         include:
           # Note: mdBook binary is incompatible with ubuntu 18.04,
           # see https://github.com/rust-lang/mdBook/issues/1779
           - os: ubuntu-18.04
-            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,valgrind,wasm-pack,wasmtime,minifier
+            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,valgrind,wasm-pack,wasmtime,cargo-watch
           - os: macos-10.15
-            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,wasm-pack,wasmtime,mdbook,minifier
+            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,wasm-pack,wasmtime,mdbook,cargo-watch
           - os: windows-2019
-            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,minifier
+            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,cargo-watch
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,11 +35,11 @@ jobs:
           # Note: mdBook binary is incompatible with ubuntu 18.04,
           # see https://github.com/rust-lang/mdBook/issues/1779
           - os: ubuntu-18.04
-            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,valgrind,wasm-pack,wasmtime
+            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,valgrind,wasm-pack,wasmtime,minifier
           - os: macos-10.15
-            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,wasm-pack,wasmtime,mdbook
+            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,wasm-pack,wasmtime,mdbook,minifier
           - os: windows-2019
-            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest
+            tool: cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,minifier
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


main.sh
```diff
@@ -30,6 +30,53 @@ warn() {
 info() {
     echo "info: $*"
 }
+cargo_binstall() {
+    tool="$1"
+    version="$2"
+
```
taiki-e, line 36: It would be nice to have info message like "install-action does not support this tool, fallback to cargo-binstall" here.

taiki-e (commented):
Thanks!

main.sh
```diff
@@ -30,6 +30,53 @@ warn() {
...
+            rm "cargo-binstall-${target}.zip"
+        else
+            retry curl --proto '=https' --tlsv1.2 -fsSL --retry 10 --retry-connrefused "$url" | tar xz
+        fi
+        mv cargo-binstall "${cargo_bin}"
+    fi
+
+    case "${version}" in
+        latest)
+            cargo binstall --no-confirm --target "$target" "$tool"
```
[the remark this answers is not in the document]
NobodyXu: There is a bug in `cargo-binstall` 0.8.0 that is going to be fixed here https://github.com/ryankurte/cargo-binstall/pull/157 and probably included in 0.8.1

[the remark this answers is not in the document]
NobodyXu: I will remove this once a new release of `cargo-binstall` that fixed this issue is published

main.sh
```diff
@@ -30,6 +30,53 @@ warn() {
...
+        x86_64-unknown-linux-gnu) url="${base_url}-x86_64-unknown-linux-musl.tgz" ;;
+        x86_64-unknown-linux-musl) url="${base_url}-x86_64-unknown-linux-musl.tgz" ;;
+
+        armv7-unknown-linux-gnueabihf) url="${base_url}-armv7-unknown-linux-musleabihf.tgz" ;;
+        armv7-unknown-linux-musleabihf) url="${base_url}-armv7-unknown-linux-musleabihf.tgz" ;;
+
+        aarch64-unknown-linux-gnu) url="${base_url}-aarch64-unknown-linux-musl.tgz" ;;
+        aarch64-unknown-linux-musl) url="${base_url}-aarch64-unknown-linux-musl.tgz" ;;
+
+        x86_64-apple-darwin | aarch64-apple-darwin | x86_64-pc-windows-msvc)
```
taiki-e, line 50: Even if the host is windows-gnu, I think we can install the windows-msvc binary because it works on the GitHub Actions' environment.

NobodyXu: Thanks, I added a new branch here for target `x86_64-pc-windows-gnu`.

main.sh
```diff
@@ -213,7 +267,7 @@ for tool in "${tools[@]}"; do
             retry curl --proto '=https' --tlsv1.2 -fsSL --retry 10 --retry-connrefused "${url}" \
                 | tar xzf - -C ${CARGO_HOME:-~/.cargo}/bin
             ;;
-        *) bail "unsupported tool '${tool}'" ;;
+        *) cargo_binstall "$tool" "$version" ;;
```
taiki-e, line 270: I think we can also support the installation of the cargo-binstall itself. It would be useful if the user wants to use non-default cargo-binstall options. (e.g., a case where the host is linux-gnu but the user wants to install linux-musl binaries.)

I think we can probably implement this by skipping the installation of `$tool` if `$tool == cargo-binstall`.
(We do not want to allow the installation of a version of cargo-binstall that is incompatible with install-action, so it is preferable not to support version specification.)

NobodyXu: That's a good idea!

I think it is better to refactor `cargo_binstall` and split out a new function `install_cargo_binstall`, as it will be much cleaner than putting it in one function.

NobodyXu: Done!

main.sh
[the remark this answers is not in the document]
NobodyXu: https://github.com/ryankurte/cargo-binstall/pull/161 and https://github.com/ryankurte/cargo-binstall/pull/157 shall fix this.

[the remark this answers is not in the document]
NobodyXu: [`cargo-binstall` v0.9.1](https://github.com/ryankurte/cargo-binstall/releases/tag/v0.9.1) has released and it not only fixed this problem, but also implement fallbacks:
 - on linux, it would first check for gnu (if present), then fallback to musl
 - on mac, it would first check for aarch64 (if on M1/M2), then fallback to x86_64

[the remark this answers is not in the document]
NobodyXu: Done!

NobodyXu:
@taiki-e I am clueless at what cause the error in the CI.

Can you help me with the error?

main.sh
```diff
@@ -213,7 +276,8 @@ for tool in "${tools[@]}"; do
             retry curl --proto '=https' --tlsv1.2 -fsSL --retry 10 --retry-connrefused "${url}" \
                 | tar xzf - -C ${CARGO_HOME:-~/.cargo}/bin
             ;;
-        *) bail "unsupported tool '${tool}'" ;;
+        cargo-binstall) install_cargo_binstall ;;
+        *) cargo_binstall "$tool" "$version" ;;
     esac
 
     info "${tool} installed at $(type -P "${tool}")"
```
taiki-e, line 283: This line and subsequent version checks will result in an error if the binary name differs from the package name. If the package was installed using cargo-binstall, these should be skipped. (Also, the cargo-binstall displays logs that are equivalent to these messages)

.github/workflows/ci.yml
```diff
@@ -26,19 +26,20 @@ jobs:
           - ubuntu-20.04
         tool:
           # Note: Specifying the version of valgrind and wasm-pack is not supported.
-          - cargo-hack,cargo-llvm-cov,cargo-minimal-versions,parse-changelog,cross,nextest,shellcheck,shfmt,valgrind,wasm-pack,wasmtime,mdbook
-          - cargo-hack@0.5.14,cargo-llvm-cov@0.4.5,cargo-minimal-versions@0.1.4,parse-changelog@0.4.8,cross@0.2.1,nextest@0.9.11,shellcheck@0.8.0,shfmt@3.5.0,wasmtime@0.37.0,mdbook@0.4.18
+          # minifier is supported by cargo-binstall (through quickinstall)
```
taiki-e, line 29: If possible, I would like to choose a crate maintained by the cargo-binstall maintainer or collaborator for testing. (maybe cargo-watch?)
This is because there is little risk of support for cargo-binstall being removed.

main.sh
```diff
@@ -30,6 +30,69 @@ warn() {
...
+        if [ $is_zip = true ]; then
+            retry curl --proto '=https' --tlsv1.2 -fsSL --retry 10 --retry-connrefused "$url" -o "cargo-binstall-${target}.zip"
+            unzip "cargo-binstall-${target}.zip"
+            rm "cargo-binstall-${target}.zip"
+        else
+            retry curl --proto '=https' --tlsv1.2 -fsSL --retry 10 --retry-connrefused "$url" | tar xz
+        fi
+
+        mkdir -p "{cargo_bin}"
+        mv cargo-binstall "${cargo_bin}"
```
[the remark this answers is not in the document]
NobodyXu: The tar/zip file of the archive contains only one file and no directory, so it should be fine.

NobodyXu:
@taiki-e I am really confused at the err msg:

```
Debugging
total 9976
   0 drwxr-xr-x  16 runner  staff   512B Jun  8 15:31 .
   0 drwxr-xr-x   3 runner  staff    96B Jun  8 15:31 ..
   8 -rw-r--r--   1 runner  staff   331B Jun  8 15:31 .editorconfig
   0 drwxr-xr-x  13 runner  staff   416B Jun  8 15:31 .git
   8 -rw-r--r--   1 runner  staff    19B Jun  8 15:31 .gitattributes
   0 drwxr-xr-x   3 runner  staff    96B Jun  8 15:31 .github
   8 -rw-r--r--   1 runner  staff   277B Jun  8 15:31 .gitignore
  16 -rw-r--r--   1 runner  staff   5.2K Jun  8 15:31 CHANGELOG.md
  24 -rw-r--r--   1 runner  staff   9.9K Jun  8 15:31 LICENSE-APACHE
   8 -rw-r--r--   1 runner  staff   1.0K Jun  8 15:31 LICENSE-MIT
  16 -rw-r--r--   1 runner  staff   6.4K Jun  8 15:31 README.md
   8 -rw-r--r--   1 runner  staff   253B Jun  8 15:31 action.yml
9848 -rwxr-xr-x   1 runner  staff   4.8M Jun  8 05:26 cargo-binstall
   8 -rw-r--r--   1 runner  staff   335B Jun  8 15:31 main.js
  24 -rwxr-xr-x   1 runner  staff    11K Jun  8 15:31 main.sh
   0 drwxr-xr-x   4 runner  staff   128B Jun  8 15:31 tools
mv: rename cargo-binstall to ~/.cargo/bin: No such file or directory
Error: Command failed: bash --noprofile --norc /Users/runner/work/install-action/install-action/main.sh
```

Clearly `cargo-binstall` is extracted correctly and the `mv` also seems to be valid for macos and windows.

taiki-e:
I believe the error is due to `~` being treated as just a string when quoted (I just remembered that I had encountered that problem before). You can remove the quote and disable SC2086 as follows, or you can use the `HOME` environment variable instead.

https://github.com/taiki-e/install-action/blob/bc0a06a003a8225fe3e896c9ed3a4c3cc2e8416a/main.sh#L80-L82

main.sh
NobodyXu: I just removed these lines.
If you think there's a better fix, please post another RFC.

taiki-e (approved):
LGTM, thanks @NobodyXu!

taiki-e:
Published in 1.4.0.

PR #9, opened 2022-06-28
Enable secure mode for `cargo-binstall`

NobodyXu:
`--secure` mode enforce downloads over secure transports only.
As a result, http will be disabled, and it will also set
min tls version to be 1.2


taiki-e (approved):
Thanks!

taiki-e:
Published in 1.5.0.

PR #16, opened 2022-07-26
Don't try to run `cargo binstall --version`

### LINKED ISSUE #14: binstall install is broken
https://github.com/CAD97/tracing-filter/runs/7512221824?check_suite_focus=true

`cargo binstall` does not have a `--version` flag to print the version, so trying to print its version fails.

https://github.com/taiki-e/install-action/blob/7e5692dc7ddd2f8da5f65e7157fcfa7f6b943093/main.sh#L349-L353

I think that switch just needs a `cargo-binstall) x cargo binstall --help | head 1` case.

CAD97:
Fixes #14

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -353,6 +353,7 @@ for tool in "${tools[@]}"; do
 
     info "${tool} installed at $(type -P "${bin}")"
     case "${bin}" in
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks.

PR #19, opened 2022-08-13
Cargo binstall force

### LINKED ISSUE #18: Use `-f` for `cargo-binstall`?
In #17, the ubuntu and windows CI failed because `$CARGO_HOME/binstall/crates-v1.json` contains the metadata of the installed crate but somehow `$CARGO_HOME/.cargo/bin` does not.

Since caching is something easy to get wrong, perhaps it is a good idea to use `-f` to force the installation.

Though `-f` and this new skip-on-installed behavior is only available from [`cargo-binstall v0.12.0`](https://github.com/cargo-bins/cargo-binstall/releases/tag/v0.12.0), install-action has to use `cargo-binstall -V` to detect the version first.

Another alternative is to upgrade `cargo-binstall` to the latest version first.

NobodyXu:
Fixed #18

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -108,7 +108,7 @@ install_cargo_binstall() {
         x cargo binstall -V
     else
         info "cargo-binstall already installed on in ${cargo_bin}/cargo-binstall, upgrading"
-        cargo binstall --secure --no-confirm  --version '>=0.12.0' cargo-binstall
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks!

taiki-e:
Published in 1.11.1.

PR #22, opened 2022-09-25
Ensure that cargo-binstall 0.13.3 is used

NobodyXu:


taiki-e:
If we do this, I think we can remove uses of `--secure` flag. (Since 0.13 that is the default, right?)

https://github.com/taiki-e/install-action/blob/618bb235a3bbe8d8a1b2d72df5f8f8ec88a76bda/main.sh#L124-L130

NobodyXu:
> (Since 0.13 that is the default, right?)

Yes

NobodyXu:
@taiki-e I've removed `--secure` flag.

taiki-e (approved):
Thanks!

taiki-e:
Hmm, it seems cargo-binstall's self-update doesn't work on windows? https://github.com/taiki-e/install-action/actions/runs/3121286524/jobs/5062493721

```
info: installing cargo-binstall@latest
info: installing cargo-binstall
info: downloading https://github.com/cargo-bins/cargo-binstall/releases/download/v0.13.3/cargo-binstall-x86_64-pc-windows-msvc.zip
Archive:  tmp.zip
  inflating: cargo-binstall.exe      
info: cargo-binstall installed at /c/Users/runneradmin/.cargo/bin/cargo-binstall.exe
+ cargo binstall -V
cargo-binstall 0.13.3

info: installing cargo-watch@latest
info: install-action does not support cargo-watch, fallback to cargo-binstall
info: cargo-binstall already installed on in /c/Users/runneradmin/.cargo/bin/cargo-binstall, upgrading
07:19:00 [INFO] Resolving package: 'cargo-binstall'
07:19:01 [INFO] The package will be downloaded from github.com
07:19:01 [INFO] This will install the following binaries:
07:19:01 [INFO]   - cargo-binstall (cargo-binstall.exe -> C:\Users\runneradmin\.cargo\bin\cargo-binstall-v0.13.3.exe)
07:19:01 [INFO] And create (or update) the following symlinks:
07:19:01 [INFO]   - cargo-binstall (C:\Users\runneradmin\.cargo\bin\cargo-binstall.exe -> C:\Users\runneradmin\.cargo\bin\cargo-binstall-v0.13.3.exe)
07:19:01 [INFO] Installing binaries...
07:19:01 [ERROR] Fatal error:
binstall::io

  × Access is denied. (os error 5)
```

from 
```yml
        uses: ./
        with:
          tool: cargo-binstall,cargo-watch
```

NobodyXu:
@taiki-e I think it might be that windows disallow removing the executable when processes of it is running?
I heard that before.

NobodyXu:
IDK, that looks strange since our [CI](https://github.com/cargo-bins/cargo-binstall/blob/c27c3b80b59d62fb23f83c4d9298070c1bfca03e/.github/scripts/tests.sh#L56) does perform upgrades as testing.

PR #24, opened 2022-09-28
Update cargo-binstall to 0.14.0

coltfred:
Update to the latest cargo-binstall, which fixes the installation of sccache.

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -80,7 +80,7 @@ host_triple() {
 }
 install_cargo_binstall() {
     # https://github.com/cargo-bins/cargo-binstall/releases
-    binstall_version="0.13.3"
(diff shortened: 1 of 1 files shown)


NobodyXu (approved):

NobodyXu:
Here's the [full changelog](https://github.com/cargo-bins/cargo-binstall/releases).

It's worth mentioning that `cargo-binstall` can also install crates like miniserve which set pkg-fmt to bin instead of packing them as a tarball/zip archive, plus some sandboxing for `bin-dir`.

taiki-e (approved):
Thanks.

taiki-e:
Published in 1.13.9.

Issue #25, opened 2022-10-02
Support partial/tilde/carret versions in `@version` syntax


Emilgardis:
Would like to be able to specify the minimal (and max) version for a tool to install.

Would enable the following

```
uses: taiki-e/install-action@v1
      with:
        tool: cargo-llvm-cov@~0.5
```

Emilgardis:
Encountered in https://github.com/cross-rs/cross/pull/1042#issuecomment-1264685777

taiki-e:
Thanks for the suggestion. Omitting the patch/minor version is already supported in some packages, and what you actually need for your use case is not to support the (complex) semver requirement, but to support omitting the patch/minor version in other packages as well, right?

https://github.com/taiki-e/install-action/blob/e85ee51dd0ff4e1362366b809fb249d9a672f562/.github/workflows/ci.yml#L32-L33

Emilgardis:
Yes! Nothing fancy needed

So this issue is only present on some of the packages, with some being those installed via `binstall` ?

NobodyXu:
`cargo-binstall` already supports this, i.e. `cargo binstall 'cargo-llvm-cov@~0.5'` or `cargo binstall --version '~0.5' cargo-llvm-cov`.

Emilgardis:
My comment was written in reverse, the issue is not present in those installed by `cargo-bininstall` I presume because of that

NobodyXu:
> My comment was written in reverse, the issue is not present in those installed by `cargo-bininstall` I presume because of that

Oh sorry :D

taiki-e:
It is the unexpected behavior that `~` is accepted when binstall is used as a fallback. If that was the intended behavior, officially supporting the package would always be a breaking change.

NobodyXu:
> officially supporting a new package would always be a breaking change

Sorry that I didn't quite understand this.
Why would supporting a new package be a breaking change?

taiki-e:
> > officially supporting a new package would always be a breaking change
> 
> Sorry that I didn't quite understand this. Why would supporting a new package be a breaking change?

1. Users install crate X using install-action's binstall fallback with `~` requirement. (install-action does not support crate X, but crate X support installation via binstall)
2. install-action now supports crate X. (No support for `~` requirements)
3. Workflow to install crate X using `~` requirement became broken because it `~` requirement is now not supported.

NobodyXu:
[...]

Issue #5, opened 2022-04-04
Auto-detecting latest tool versions


thomcc:
It seems like the list of versions is manually maintained. This seems hard to scale and already is at the point where it would be hard to blame you for failing to notice that a release happened.

I'm hoping to use `install-action` to replace some similar code (in an unreleased project). However, remembering to update to the latest version seemed impractical to me, so my code auto-detected it using the github API + `jq` (`jq` is pre-installed on GHA runners). The code was something like this:

```sh
api_url="https://api.github.com/repos/$repo/releases/latest"
query='.assets[] | select(.name | endswith("x86_64-unknown-linux-gnu.tar.gz")) | .browser_download_url'

tarball_url="$(curl --proto '=https' --tlsv1.2 --retry 5 --retry-delay 10 -sfSL "$api_url" | jq -r "$query")"
# then fetch the tarball was fetched with 
# curl <same flags> "$tarball_url" | tar zxf -
```

This hard-codes several things which would need to not be hard-coded for `install-action` usage, but it may be useful as a starting point. (Also, I'm not good at `sh`, so this may do something that is not ideal in some way). But you get the picture.

Although, maybe it's not how you'd like to do it. There's an argument that auto-detecting the latest version in this way is fragile, both because it makes too many API requests, and because it breaks if a release happens without binaries. It also doesn't easily support usage with a checksum as required by #1.

So perhaps it would be better to instead automate the updates of `install-action`'s version list in some manner. This seems harder to me though, but perhaps you have plans.

Anyway, sorry if my concern is misplaced and you do not think this will be a problem. I just felt a little bit bad adding to your workload by adding a new tool after I noticed that versions were hard-coded.

Thank you.

taiki-e:
Thanks for writing this! I strongly agree that auto-detection of the latest version is necessary.
(Currently, I track the latest releases by watching the releases of these repositories, but I acknowledge that it is an easy-to-forget and non-scalable way.)

> There's an argument that auto-detecting the latest version in this way is fragile, both because it makes too many API requests, and because it breaks if a release happens without binaries.

Yes. In the past, this action used the similar way that you mentioned (https://github.com/taiki-e/install-action/commit/56a03286dbf2186895b23d8966c0998292e19af6), but I have encountered both of these problems.

> So perhaps it would be better to instead automate the updates of install-action's version list in some manner. This seems harder to me though, but perhaps you have plans.

[...]

Issue #1, opened 2021-12-30
Checksum


taiki-e:
The current installation way is secure enough.

> When installing the tool from GitHub Releases, this action will download the tool or its installer from GitHub Releases using HTTPS with tlsv1.2+. This is basically considered to be the same level of security as [the recommended installation of rustup](https://www.rust-lang.org/tools/install).

However, verifying the hashes of downloaded files can provide a higher level of security.

> Cases where the current installation way is vulnerable and checksum can prevent:
>
> 1. The GitHub account is hijacked and the release binary is replaced with another malicious program.
> 1. After a GitHub account is renamed, a GitHub account, project, and releases with the same name are created by a malicious user.
 
However, the checksum is not perfect, as it cannot reject new releases on hijacked accounts.

Perhaps the best approach here from a security standpoint is to verify the signature of the binary or archive, but the compiled binaries of most projects we support are not signed. Also, depending on how it is signed, it may be vulnerable to hijacking in any case.

PR #26, opened 2022-10-03
Fix `main.sh`: bail on semver operators

NobodyXu:

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -153,7 +153,7 @@ fi
 for tool in "${tools[@]}"; do
     if [[ "${tool}" == *"@"* ]]; then
         version="${tool#*@}"
-        if [[ ! "${version}" =~ ^[1-9][0-9]*$|^[0-9]+\.[1-9][0-9]*$|^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?$|^latest$ ]]; then
+        if [[ ! "${version}" =~ ^([1-9][0-9]*(\.[0-9]+(\.[0-9]+)?)?|0\.[1-9][0-9]*(\.[0-9]+)?|^0\.0\.[0-9]+)(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?$|^latest$ ]]; then
             bail "takie-e/install-action does not support semver operators"
         fi
     else
--- a/main.sh
+++ b/main.sh
@@ -153,7 +153,7 @@ fi
 for tool in "${tools[@]}"; do
     if [[ "${tool}" == *"@"* ]]; then
         version="${tool#*@}"
-        if [[ "${version}" != "latest" ]] && ! [[ "${version}" =~ ^[0-9]*\.[0-9]*(\.[0-9]*)?$ ]]; then
+        if [[ ! "${version}" =~ ^[1-9][0-9]*$|^[0-9]+\.[1-9][0-9]*$|^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?$|^latest$ ]]; then
             bail "takie-e/install-action does not support semver operators"
         fi
     else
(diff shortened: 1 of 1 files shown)


main.sh
```diff
@@ -153,6 +153,9 @@ fi
 for tool in "${tools[@]}"; do
     if [[ "${tool}" == *"@"* ]]; then
         version="${tool#*@}"
+        if [[ "${version}" != "latest" ]] && ! [[ "${version}" =~ ^[0-9]*\.[0-9]*(\.[0-9]*)?$ ]]; then
```
taiki-e, line 156: Well, the semver version is actually a bit more complicated because it allows prerelease and metadata.
FYI, if we accept only the exact valid semver version, probably then:

```sh
if [[ ! "${version}" =~ ^[1-9][0-9]*(\.(0|[1-9][0-9]*)(\.(0|[1-9][0-9]*)(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?)?)?$|^0\.[1-9][0-9]*(\.(0|[1-9][0-9]*)(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?)?$|^0\.0\.(0|[1-9][0-9]*)(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?$|^latest$ ]]; then
```

However, we don't need something so exacting, so the following should probably be fine:


```suggestion
        if [[ ! "${version}" =~ ^[1-9][0-9]*$|^[0-9]+\.[1-9][0-9]*$|^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?$|^latest$ ]]; then
```

taiki-e (changes requested):
Thanks!

main.sh
```diff
@@ -153,6 +153,9 @@ fi
 for tool in "${tools[@]}"; do
     if [[ "${tool}" == *"@"* ]]; then
         version="${tool#*@}"
+        if [[ ! "${version}" =~ ^[1-9][0-9]*$|^[0-9]+\.[1-9][0-9]*$|^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?$|^latest$ ]]; then
```
taiki-e, line 156: Sorry, I noticed that the code I previously suggested could not handle a version like `1.0`...

```suggestion
        if [[ ! "${version}" =~ ^([1-9][0-9]*(\.[0-9]+(\.[0-9]+)?)?|0\.[1-9][0-9]*(\.[0-9]+)?|^0\.0\.[0-9]+)(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?$|^latest$ ]]; then
```

NobodyXu: It's getting so complicated so quickly...

Maybe we should rewrite in python?
Then we can also use batch installation mode of `cargo-binstall`.

taiki-e:
> It's getting so complicated so quickly...

Well, we can also reduce complexity by accepting some odd versions like `0` or `0.0`. (I'm fine with either.)

```sh
if [[ ! "${version}" =~ ^[0-9]+(\.[0-9]+(\.[0-9]+)?)?(-[0-9A-Za-z\.-]+)?(\+[0-9A-Za-z\.-]+)?$|^latest$ ]]; then
```


> Maybe we should rewrite in python?

I'm not very familiar with python, so there is no possibility to do that in this repository.

--- 

Btw, as for reducing the complexity of install-action, improving maintainability, and support for #1, #5, #25, etc., I'm thinking of managing the package information as json or another and fetching the latest version's package information in CI to automatically update (like my other repos I mentioned in https://github.com/taiki-e/install-action/issues/5#issuecomment-1087198391).

NobodyXu:
> Well, we can also reduce complexity by accepting some odd versions like 0 or 0.0. (I'm fine with either.)

I'm fine with this PR.

> I'm thinking of managing the package information as json or another and fetching the latest version's package information in CI to automatically update (like my other repos I mentioned in https://github.com/taiki-e/install-action/issues/5#issuecomment-1087198391).

That will definitely makes updating a lot easier.

PR #34, opened 2022-12-24
Support just and dprint

### LINKED ISSUE #33: Support just and dprint
Hi.

I have been using `install-action` across my project's CI. Most of them are `just` and `dprint`.

Would you like to support `just` and `dprint`? 
Is falling back to `binstall` took longer than built-in support crates?

Thanks a lot for insall-action :heart:

taiki-e:
Closes #33 

- [just](https://github.com/casey/just) 1.9.0
- [dprint](https://github.com/dprint/dprint) 0.34.1

cc @azzamsa


tools/codegen/base/dprint.json
```diff
@@ -0,0 +1,13 @@
+{
+  "repository": "https://github.com/dprint/dprint",
+  "tag_prefix": "",
+  "asset_name": "${package}-${rust_target}.zip",
+  "version_range": ">= 0.34.1",
```
taiki-e, line 5: Hmm. As for tools after #27, we would like to support basically the latest release at the time of addition and later versions to reduce the manifest size bloat. However, this appears to cause problems similar to https://github.com/taiki-e/install-action/issues/25#issuecomment-1265219010...

taiki-e: Added manifest for older versions for now. As for the manifest size issue, I will track it in https://github.com/taiki-e/install-action/issues/36.

azzamsa:
Any task I can help to get it merged?

PR #54, opened 2023-01-15
Use cargo-binstall batching mode

### LINKED ISSUE #44: Use `cargo-binstall` batching mode?
`cargo-binstall` can install multiple crates in one invocation:

```
cargo binstall cargo-nextest cargo-binstall cargo-release
```

which would download, extract and install the crates in parallel.

I'm not sure how beneficial that would be, but surely it would speed up CI if it installs multiple tools or one of the tools needs to download from a relatively slow server.

taiki-e:
Closes #44

### DIFF
--- a/.github/.cspell/project-dictionary.txt
+++ b/.github/.cspell/project-dictionary.txt
@@ -26,4 +26,5 @@ shellcheck
 shfmt
 udeps
 wasmtime
+watchexec
 xscale
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,21 +29,21 @@ jobs:
           - ubuntu-20.04
           - ubuntu-22.04
         tool:
-          # cargo-watch is supported by cargo-binstall (through quickinstall)
+          # cargo-watch/watchexec-cli is supported by cargo-binstall (through quickinstall)
           # TODO: valgrind installation sometime hangs.
-          - cargo-hack,cargo-llvm-cov,cargo-minimal-versions,cargo-no-dev-deps,parse-changelog,cargo-udeps,cargo-valgrind,cargo-deny,cross,dprint,just,nextest,protoc,shellcheck,shfmt,wasm-pack,wasmtime,mdbook,mdbook-linkcheck,cargo-watch,grcov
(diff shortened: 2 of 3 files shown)


taiki-e:
hmm https://github.com/taiki-e/install-action/actions/runs/3923398658/jobs/6706983450 
```
WARN The package watchexec-cli v1.21.0 will be installed from source (with cargo)
```

NobodyXu:
> hmm https://github.com/taiki-e/install-action/actions/runs/3923398658/jobs/6706983450
> 
> ```
> WARN The package watchexec-cli v1.21.0 will be installed from source (with cargo)
> ```

That could be switched off using `--disable-strategies compile`

NobodyXu:
@taiki-e I tried installing this from my own laptop and it works fine:

```
❯ cargo binstall watchexec-cli --targets x86_64-unknown-linux-musl
 INFO resolve: Resolving package: 'watchexec-cli'
 WARN The package watchexec-cli v1.21.0 will be downloaded from github.com
 INFO This will install the following binaries:
 INFO   - watchexec (watchexec -> /Users/nobodyxu/.cargo/bin/watchexec-v1.21.0)
 INFO And create (or update) the following symlinks:
 INFO   - watchexec (/Users/nobodyxu/.cargo/bin/watchexec -> watchexec-v1.21.0)
Do you wish to continue? yes/[no]
? no
 WARN Installation cancelled
```

taiki-e:
> That could be switched off using `--disable-strategies compile`

Hmm, it would be a breaking change...

That said, if installing from source, it is better to cache the binaries, so the fallback that installs from source is not very good to me.

So I'm considering outputting info message suggesting [cache-cargo-install-action](https://github.com/taiki-e/cache-cargo-install-action) in all officially unsupported cases. (Originally, I was considering outputting it when the installation fails)

Ideally, I would like to merge cache-cargo-install-action into install-action, run cargo-binstall with `--disable-strategies compile`, and use cache-cargo-install-action when the installation fails, though...

NobodyXu:
> Hmm, it would be a breaking change...

Yes, so better to fix it in cargo-binstall.

> Ideally, I would like to merge cache-cargo-install-action into install-action, run cargo-binstall with --disable-strategies compile, and use cache-cargo-install-action when the installation fails, though...

That sounds awesome!

PR #58, opened 2023-01-16
Switch to composite action

taiki-e:
This switching is needed to implement the idea said in https://github.com/taiki-e/install-action/pull/54#issuecomment-1383149339. (To actually do it, we also need to address the cache-cargo-install-action's limitation that [cannot specify multiple crates in a single action call](https://github.com/taiki-e/cache-cargo-install-action#migrate-fromto-install-action), though.)

> Ideally, I would like to merge cache-cargo-install-action into install-action, run cargo-binstall with `--disable-strategies compile`, and use cache-cargo-install-action when the installation fails, though...

I tried this twice in the past and gave up due to https://github.com/actions/runner/issues/665 and https://github.com/actions/runner/issues/2185, respectively, and eventually was able to work around these problems when implementing cache-cargo-install-action.


taiki-e:
A potential concern regarding this is compatibility with older self-hosted runners.

Composite action [was added on 2020-08-07](https://github.blog/changelog/2020-08-07-github-actions-composite-run-steps), but it probably requires [2.290.0](https://github.com/actions/runner/releases/tag/v2.290.0) (2022-04-13) or [2.295.0](https://github.com/actions/runner/releases/tag/v2.295.0) (2022-08-10) for this action to actually function properly in the composite action of the self-hosted runners.

taiki-e:
Given that it has been 6 months since the release of runner, which is expected to work, and that node 16 will be EoL in about 6 more months, I think this is a reasonable time to merge this.

PR #65, opened 2023-02-06
Add `cargo-tarpaulin` support

orhun:
See #42


tools/codegen/base/cargo-tarpaulin.json
```diff
@@ -0,0 +1,19 @@
...
+  "tag_prefix": "",
+  "asset_name": "${package}-${rust_target}.tar.gz",
+  "version_range": "> 0.22.0",
+  "platform": {
+    "x86_64_linux_gnu": {},
+    "x86_64_linux_musl": {},
+    "aarch64_linux_gnu": {},
+    "aarch64_linux_musl": {},
+    "x86_64_macos": {
+      "asset_name": "${package}-universal-apple-darwin.tar.gz"
```
taiki-e, line 12: Why not use x86_64 binaries?

taiki-e (commented):
Thanks!

tools/codegen/base/cargo-tarpaulin.json
orhun: Not sure, took that from here:

https://github.com/taiki-e/install-action/blob/f30d182e9eb6ec2e7e967394d0db4b42f5984431/tools/codegen/base/cargo-nextest.json#L8-L10

I can change it if you want 🐻

orhun: PR is updated!

taiki-e:
Published in 2.4.0.

PR #71, opened 2023-02-10
Add support for zola

### LINKED ISSUE #70: Add support for zola
I'd like to request support for https://github.com/getzola/zola. It's a static website generator with a solid usecase in github actions.

tranzystorek-io:
Closes #70

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,19 +40,19 @@ jobs:
         tool:
           # cargo-watch/watchexec-cli is supported by cargo-binstall (through quickinstall)
           # TODO: valgrind installation sometime hangs.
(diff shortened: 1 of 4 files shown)


taiki-e:
Thanks! Could you add zola to this line as well? 
https://github.com/taiki-e/install-action/blob/6e270f6359e00e6e4b83c8edc6d769c1947b7f5b/.github/workflows/ci.yml#L102-L104

tranzystorek-io:
It fails the test on some of those containers if I add it there: https://github.com/taiki-e/install-action/actions/runs/4148390689

taiki-e:
Oh, in that case, it is ok as is. (I will need to split the matrix into a glibc 2.31+ group and a 2.27+ group later.)

taiki-e (approved):

taiki-e:
Published in 2.4.0.

PR #124, opened 2023-05-29
Pass --locked to cargo-binstall

NobodyXu:
So that if taiki-e/install-action fallback to cargo-install, it will use the lockfile instead of latest version for every dep to ensure the compilation succeeds.


taiki-e (approved):
Thanks!

PR #136, opened 2023-06-10
Propose to put each tested tools in the CI on a new line

FirelightFlagboy:
I find it more simple to see which tools are tested a which step.


taiki-e:
Thanks for the PR. I'm concerned that this will make CI files bloated and rather difficult to read.

These lists are the same on most platforms except for a few, so I would prefer to generate them automatically with scripts. (#137)

FirelightFlagboy:
I understand, generating the list of tests for the CI like proposed in #137 would be a good improvement.

So I will close this pull-request

Issue #178, opened 2023-08-02
Update nextest to 0.9.57


sunshowers:
Hi @taiki-e -- could you update nextest to 0.9.57? 0.9.56 has a bad bug and I needed to yank it. Thanks!

sunshowers:
(Separately, I'd like us to also figure out a way to quickly yank bad versions from the install-action.)

NobodyXu:
Update is done, just waiting for a new release.

NobodyXu:
P.S. all updates are automatic created by GHA, so there's no need to opening an issue manually since we won't miss it.

You can have relax knowning that this repository will always keep them up to date!

Thank you for openin this issue.

sunshowers:
@NobodyXu, thanks. I think it's just a matter of urgency -- once an update has been yanked on crates.io (which was done within 5-10 minutes), it would be nice if the action no longer tried to fetch that version.

NobodyXu:
> @NobodyXu, thanks. I think it's just a matter of urgency -- once an update has been yanked on crates.io (which was done within 5-10 minutes), it would be nice if the action no longer tried to fetch that version.

Unfortunately the latest version info and checksum is hardcoded into the crate, so it requires a new release.

NobodyXu:
In terms of this, cargo-binstall can indeed fetch the latest version since it does not hardcode anything.

taiki-e:
Published in 2.13.5.

sunshowers:
Thanks for all your hard work maintaining the action, Taiki!

taiki-e:
> way to quickly yank bad versions from the install-action

I think it is possible to fetch the crates.io API during manifest generation to prevent the yanked version from being selected as a candidate for the latest or omitted version. This is not likely to very help this case, where the problem was resolved very quickly on the nextest side, since it is not reflected until the manifest update is released, but it would be helpful if the fix is not released quickly. It will also make the way described below more efficient.

I think it is also possible to fetch the crates.io API during installation and install the older version if the version is yanked (unless that version is explicitly specified). This could be implemented by adding a field to the manifest that makes fetching the older version easier, such as "previous_stable_version", and a field that indicates whether we want to fetch crates.io's API, such as "is_rust_crate", and code on the action side to fetch crates.io's API and select the version.

NobodyXu:
> > way to quickly yank bad versions from the install-action
> 
[...]

PR #182, opened 2023-08-04
Fetch crates.io API during installation and install older version if latest version is yanked

taiki-e:
Implement the second one in https://github.com/taiki-e/install-action/issues/178#issuecomment-1663273376

> I think it is also possible to fetch the crates.io API during installation and install the older version if the version is yanked (unless that version is explicitly specified). This could be implemented by adding a field to the manifest that makes fetching the older version easier, such as "previous_stable_version", and a field that indicates whether we want to fetch crates.io's API, such as "is_rust_crate", and code on the action side to fetch crates.io's API and select the version.

The first one is already implemented in #180.

cc @sunshowers

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -163,7 +163,24 @@ read_manifest() {
         exact_version="${version}"
     else
         manifest=$(jq -r ".\"${exact_version}\"" "${manifest_dir}/${tool}.json")
+        if [[ "${rust_crate}" != "null" ]]; then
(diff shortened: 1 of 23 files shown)


taiki-e:
This is simple enough and already complete, but since the maintainer of nextest, the only crate that actually needs this process, seems to prefer to use cargo-binstall, let's discard this in favor of a patch to make it use cargo-binstall.

NobodyXu (approved):

PR #183, opened 2023-08-04
Use cargo-binstall to install nextest

taiki-e:
Replaces #182

As mentioned in https://github.com/taiki-e/install-action/pull/182#issuecomment-1665600245, the maintainer of nextest [seems to prefer to use cargo-binstall](https://github.com/taiki-e/install-action/issues/178#issuecomment-1665118419).

Internally, this is handled in a similar way to the unsupported tool, but the `@nextest`/`@cargo-nextest` alias/shorthand continues to be supported.


taiki-e:
cc @sunshowers

NobodyXu (approved):

sunshowers:
Looks good, thanks!

sunshowers:
(Looks like it does.)

NobodyXu:
> @NobodyXu will specifying a version of "0.9" continue to work with binstall?

Yes, `cargo-binstall` uses [`semver`](https://docs.rs/semver) to find the latest version on crates.io that is not yanked.

You can also pass `>=0.9`, `<=0.9`, `>=0.9.1,<=0.9.3`

taiki-e:
Published in 2.14.0.

PR #212, opened 2023-08-29
Add support for xbuild

notgull:
xbuild is a tool for building iOS/Android apps in Rust.

### DIFF
--- a/manifests/xbuild.json
+++ b/manifests/xbuild.json
@@ -1,0 +1,34 @@
+{
+  "rust_crate": "xbuild",
+  "template": {
+    "x86_64_linux_musl": {
+      "url": "https://github.com/rust-mobile/xbuild/releases/download/v${version}/xbuild-linux-x64",
+      "bin": "xbuild-linux-x64"
+    },
+    "x86_64_macos": {
+      "url": "https://github.com/rust-mobile/xbuild/releases/download/v${version}/xbuild-macos-x64",
+      "bin": "xbuild-macos-x64"
+    },
+    "x86_64_windows": {
+      "url": "https://github.com/rust-mobile/xbuild/releases/download/v${version}/xbuild-windows-x64.exe",
+      "bin": "xbuild-windows-x64.exe"
+    }
+  },
+  "latest": {
+    "version": "0.2.0"
+  },
+  "0.2": {
+    "version": "0.2.0"
+  },
+  "0.2.0": {
+    "x86_64_linux_musl": {
+      "checksum": "9c294809ec3cb314b34e9f644b5bbf6ed262c25c1eeb78a3691836a95bc58e0d"
+    },
+    "x86_64_macos": {
+      "checksum": "e357205fde5262d77b8b1dc1cb066a4f7b4edd0a83b71b3513ae466b08343305"
+    },
+    "x86_64_windows": {
+      "checksum": "ce503cd64e6449d1c3e3bc2bfb621e9dd3ebfc775e99c97f98c2ab145e533946"
+    }
+  }
+}
--- a/tools/codegen/base/xbuild.json
+++ b/tools/codegen/base/xbuild.json
@@ -1,0 +1,12 @@
+{
+  "repository": "https://github.com/rust-mobile/xbuild",
+  "tag_prefix": "v",
+  "rust_crate": "${package}",
+  "asset_name": "${package}-${os_name}-x64${exe}",
+  "bin": "${package}-${os_name}-x64${exe}",
+  "platform": {
+    "x86_64_linux_musl": {},
+    "x86_64_macos": {},
+    "x86_64_windows": {}
+  }
+}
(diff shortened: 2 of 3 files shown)


taiki-e:
https://github.com/taiki-e/install-action/actions/runs/6014566895/job/16314554693

> xbuild tool is deprecated and will be removed in future updates, use msbuild instead

🤔

notgull:
This is odd. The maintainer for `rust-mobile` just told me [`cargo-apk` is deprecated in favor of `xbuild`](https://github.com/rust-mobile/cargo-apk/issues/31#issuecomment-1696820576).

notgull:
Oh wait, that's Mono's `xbuild` command.

It looks like I should install the `xbuild` binary under the name `x`, are there any existing controls for doing that?

taiki-e:
You can add a special case here.

https://github.com/taiki-e/install-action/blob/cc5a5c56a296ea597e6ea38f551f25dee8be1225/main.sh#L72

Probably something like.

```sh
case "${tool}" in
    build) installed_bin="${bin_dir}/x"  ;;
    *) installed_bin="${bin_dir}/$(basename "${bin_in_archive}")"  ;;
esac
```

notgull:
Ugggh, `xbuild` dynamically links to OpenSSL, doesn't it.

I've opened https://github.com/rust-mobile/xbuild/issues/132 to deal with this. Until then I guess I'll just install `libssl`.

notgull:
Hmm, it looks like `xbuild` needs SSL 3, and many of the operating systems in the configuration at the moment just don't have SSL 3 available.

taiki-e:
For now, I think we can adjust https://github.com/taiki-e/install-action/blob/main/tools/ci/tool-list.sh and mark xbuild as glibc_pre_2_34_incompat.

> many of the operating systems in the configuration

We test a number of environments in CI, but the ones that actually important for most users are the -latest variants of the GitHub-hosted runner: ubuntu-latest (ubuntu-22.04), macos-latest (macos-12), and windows-latest (windows-2022)

tools/codegen/base/xbuild.json
```diff
@@ -0,0 +1,12 @@
+{
+  "repository": "https://github.com/rust-mobile/xbuild",
+  "tag_prefix": "v",
+  "rust_crate": "${package}",
+  "asset_name": "${package}-${os_name}-x64${exe}",
+  "bin": "${package}-${os_name}-x64${exe}",
+  "platform": {
+    "x86_64_linux_musl": {},
```
taiki-e, line 8: IIUC, this is actually not linux-musl, is linux-gnu.
https://github.com/rust-mobile/xbuild/blob/c2f3291a02e8dddf531b2583f8e15b82f1461c61/.github/workflows/release.yml#L13

main.sh
```diff
@@ -565,6 +570,15 @@ for tool in "${tools[@]}"; do
...
                     ;;
+                xbuild)
+                    case "${host_os}" in
+                        linux)
+                            if type -P xbuild &>/dev/null; then
+                                sys_install libssl-dev
+                            fi
+                            ;;
+                    esac
+                    ;;
```
taiki-e, line 581: This isn't actually necessary, right? (libssl.so.N is a requirement for curl.)

taiki-e:
At least in their 0.1.x versions, the release tag does not seem to match the crate version.

https://crates.io/crates/xbuild/versions
https://github.com/rust-mobile/xbuild/releases

If that has been fixed in 0.2.x and up, that is fine, but otherwise it will be confusing because it will install something that does not match the crate version specified by the user.

notgull:
Hmm, yes, that seems problematic. 

@dvc94ch When you make `xbuild` releases from this point on, can you make sure that the release tag matches the crate version?

manifests/xbuild.json
```diff
@@ -0,0 +1,44 @@
...
+    "x86_64_macos": {
+      "url": "https://github.com/rust-mobile/xbuild/releases/download/v0.1.0%2B4/xbuild-macos-x64",
+      "checksum": "846c704d512912a55dbe1fdff8d50d6408a17a557931ed6d5fa070fc24fbf370",
+      "bin": "xbuild-macos-x64"
+    },
+    "x86_64_windows": {
+      "url": "https://github.com/rust-mobile/xbuild/releases/download/v0.1.0%2B4/xbuild-windows-x64.exe",
+      "checksum": "e24df52574fbba3b707ec85502f7515a78401cf06bae322204def565f62f555c",
+      "bin": "xbuild-windows-x64.exe"
+    }
```
taiki-e, line 42: (version_range doesn't affect already generated manifests. -- It is a design to handle the case where the URL has changed.)

```suggestion
```

dvc94ch:
We already have a gh action: https://github.com/rust-mobile/setup-xbuild-action

notgull:
> We already have a gh action: https://github.com/rust-mobile/setup-xbuild-action

It looks like that uses the GitHub CLI though, which needs a secret token to work. `install-action` works without any GitHub tokens.

notgull:
@dvc94ch Poke. Given the reasons above I feel that this is a good reason why `xbuild` should support `install-action` through proper versioning.

dvc94ch:
Yeah, really annoying that gh cli requires a token to download a release, I know. xbuild has different components I think the release tag versions the sdk build and has nothing to do with the crate version

notgull:
> Yeah, really annoying that gh cli requires a token to download a release, I know. xbuild has different components I think the release tag versions the sdk build and has nothing to do with the crate version

Hmm, so should this crate just ignore versions with `+` in it?

taiki-e (approved):
Thanks!

taiki-e:
Published in 2.18.0.

PR #228, opened 2023-09-17
Update manifest

taiki-e:
Auto-generated by [create-pull-request](https://github.com/peter-evans/create-pull-request)

PR #237, opened 2023-09-27
Support signature verification (minisign)

taiki-e:
This supports signature verification for the following format:

algorithm: minisign
public key: `package.metadata.binstall.signing.pubkey` at Cargo.toml

This is the format that cargo-binstall supports: <https://github.com/cargo-bins/cargo-binstall/blob/HEAD/SIGNING.md>

It should be easy to support public key in other places, but I support only what we need for now.

The implementation is basically the same as that described in https://github.com/taiki-e/install-action/pull/228#issuecomment-1722477611.

> We can extend the codegen to ensure that the hash in the generated manifest is the same as that obtained from the signed archive or indicated in the signed checksum file. This allows signature verification to be performed without additional cost at the time of execution of the action.

AFAIK, currently, there are no tools other than cargo-binstall that distribute signed archives in the tools we support.


taiki-e:
https://github.com/taiki-e/install-action/actions/runs/6326816086/job/17181208549

```
downloading releases of https://github.com/cargo-bins/cargo-binstall from https://api.github.com/repos/cargo-bins/cargo-binstall/releases
downloading crate info from https://crates.io/api/v1/crates/cargo-binstall
update manifest for versions '=1.4.1'
downloading https://github.com/cargo-bins/cargo-binstall/releases/download/v1.4.1/cargo-binstall-x86_64-unknown-linux-musl.tgz for checksum ... download complete
getting sha256 hash for https://github.com/cargo-bins/cargo-binstall/releases/download/v1.4.1/cargo-binstall-x86_64-unknown-linux-musl.tgz
382c45dfbdbcaee37f10cb37d6e519c75d174622cab9e287521eaa3b5f71e162 *cargo-binstall-x86_64-unknown-linux-musl.tgz
downloading https://github.com/cargo-bins/cargo-binstall/releases/download/v1.4.1/cargo-binstall-x86_64-unknown-linux-musl.tgz.sig for signature validation ... download complete
downloading https://crates.io/api/v1/crates/cargo-binstall/1.4.1/download for signature verification ... download complete
algorithm: minisign
pubkey: RWQuEpy2Ocjox5ppbdesj/VYAKvGyXITtw8OelPN1xY40izwAG/wSMvg
verifying signature for https://github.com/cargo-bins/cargo-binstall/releases/download/v1.4.1/cargo-binstall-x86_64-unknown-linux-musl.tgz ... done
```

NobodyXu (approved):

PR #246, opened 2023-10-05
feat: add biome

### LINKED ISSUE #244: Support for Biome
[Biome](https://biomejs.dev) is a linter/formatter for the web that is also written in Rust. It also has releases available on GitHub: https://github.com/biomejs/biome/releases

I think it would be nice if Biome could be installed through `install-action`, especially since it's a convenient tool to invoke from a `justfile` for which you also provide support.

cc @ematipico

nhedger:
Hey 👋,

This PR adds `biome`, as discussed in #244.

---

Closes #244

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ https://spdx.org/licenses
 
 | Name | Where binaries will be installed | Where will it be installed from | Supported platform | License |
 | ---- | -------------------------------- | ------------------------------- | ------------------ | ------- |
+| [**biome**](https://biomejs.dev)| `/usr/local/bin` | [GitHub Releases](https://github.com/biomejs/biome/releases) | Linux, macOS, Windows | [MIT](https://github.com/biomejs/biome/blob/main/LICENSE)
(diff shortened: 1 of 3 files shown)


nhedger:
I believe the tests are failing because the biome CLI currently exits with `1` when running `biome --help` or `biome --version`, which is something we fixed on main, but not in the current release.

taiki-e:
Thanks for the PR!

> If there is no way around that, perhaps we should wait until the next release.

You can also handle biome as a special case [here](https://github.com/taiki-e/install-action/blob/739d45026e01f3797a1dee51ce205a92cec535e0/main.sh#L667) (by adding `biome) rx "${tool_bin}" --version || true ;;`).

nhedger:
Thank you, @taiki-e, works like a charm!

taiki-e (approved):

taiki-e:
Published in 2.20.0.

nhedger:
Thanks for merging and your help @taiki-e !

arendjr:
Thanks for picking this up @nhedger !

PR #265, opened 2023-10-27
Add cargo-rdme to supported tools

joshka:


joshka:
The change in https://github.com/taiki-e/install-action/pull/265/commits/6b8955b3299acaef59f0307929e4d6eb7947375c uses the release zip files instead of the zst files (which CI seems to choke on)

taiki-e:
> zst files

To use tar.zst, you have to add `*.tar.zst` branch to [this `case` block](https://github.com/taiki-e/install-action/blob/0e17f1a03533b48d71f4c916c50cf16ab2f3cd33/main.sh#L72).

joshka:
> > zst files
> 
> To use tar.zst, you have to add `*.tar.zst` case to [this `case` block](https://github.com/taiki-e/install-action/blob/0e17f1a03533b48d71f4c916c50cf16ab2f3cd33/main.sh#L72).

I noticed that - figured I'd try to avoid it instead, not sure why the bin location was broken in the previous CI run, but figure it's better to just add zstd instead.

main.sh
```diff
@@ -100,6 +100,18 @@ download_and_extract() {
                 esac
             fi
             ;;
+        *.tar.zst | *.tzst)
+            tar_args+=("xvf" "--zstd")
```
taiki-e, line 104:
```suggestion
            tar_args+=("--zstd" "xvf")
```

(Argument after f (-f) is considered filename.)

joshka: :facepalm: of course :D

joshka:
Whew - got some green now at least :D Still failing a few OSes though.

joshka:
Removed the zstd code as it was incompatible with macos-11, alpine, almalinux, centos (using bz2 and zip files from cargo-rdme instead). I'm sure there's a way to get that going, but my goal was not to make zstd work for install-action, it was to just get cargo-rdme available as a tool for CI.

Last change was fixing the tidy dictionary for rdme.

joshka:
Finally all the checks are passing on this (I love when CI servers make me feel like the biggest noob :)) I've rebased the PR on main in the last push to make it easy to merge.

taiki-e (approved):
Thanks!

taiki-e:
Published in 2.21.0.

Issue #298, opened 2023-12-10
codegen fails due to missing Cargo.toml package.version field


jayvdb:
This occurs on latest stable

```
install-action> ./tools/manifest.sh osv-scanner
error: failed to load manifest for workspace member `/home/jayvdb/actions/install-action/tools/codegen`

Caused by:
  failed to parse manifest at `/home/jayvdb/actions/install-action/tools/codegen/Cargo.toml`

Caused by:
  TOML parse error at line 1, column 1
    |
  1 | [package]
    | ^^^^^^^^^
  missing field `version`
```

NobodyXu:
Perhaps setting `publish = false` will fix this?

jayvdb:
:+1: I'll try that

jayvdb:
Adding `publish = false` doesnt help.

taiki-e:
The latest cargo doesn't require package.version field: https://github.com/rust-lang/cargo/pull/12786

However, that does not seem to be included yet in the current stable and will be included in 1.75, which will be released later this month.

PR #299, opened 2023-12-10
Add package.version to Cargo.toml

jayvdb:
Closes https://github.com/taiki-e/install-action/issues/298


jayvdb:
Comments on issue explain why this wont be necessary in the next cargo stable

Issue #81, opened 2023-03-02
Add support for osv-scanner


jayvdb:
Go devsecops tool, which has binary releases at https://github.com/google/osv-scanner/releases

I'm happy to create a PR if there is appetite for this addition.

taiki-e:
I would accept a PR to support this.

PR #300, opened 2023-12-10
Add osv-scanner

jayvdb:
Closes https://github.com/taiki-e/install-action/issues/81


NobodyXu (dismissed):
Can you update the CI to test this please?

jayvdb:
Note https://github.com/taiki-e/install-action/blob/main/DEVELOPMENT.md doesnt mention adding anything to CI to test new tool additions, so I am not sure if what I've done is the appropriate way to do that.

README.md
```diff
@@ -103,6 +103,7 @@ https://spdx.org/licenses
 | [**mdbook-linkcheck**](https://github.com/Michael-F-Bryan/mdbook-linkcheck) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/Michael-F-Bryan/mdbook-linkcheck/releases) | Linux, macOS, Windows | [MIT](https://github.com/Michael-F-Bryan/mdbook-linkcheck/blob/HEAD/LICENSE) |
 | [**mdbook**](https://github.com/rust-lang/mdBook) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/rust-lang/mdBook/releases) | Linux, macOS, Windows | [MPL-2.0](https://github.com/rust-lang/mdBook/blob/HEAD/LICENSE) |
 | [**nextest**](https://github.com/nextest-rs/nextest) (alias: `cargo-nextest`) | `$CARGO_HOME/bin` | cargo-binstall | Linux, macOS, Windows | [Apache-2.0](https://github.com/nextest-rs/nextest/blob/HEAD/LICENSE-APACHE) OR [MIT](https://github.com/nextest-rs/nextest/blob/HEAD/LICENSE-MIT) |
+| [**osv-scanner**]( https://github.com/google/osv-scanner) | `$HOME/.install-action/bin` | [GitHub Releases]( https://github.com/google/osv-scanner/releases) | Linux, macOS, Windows | [Apache-2](https://github.com/google/osv-scanner/blob/main/LICENSE) |
```
jayvdb, line 106: I copied `$HOME/.install-action/bin` from the `protoc` entry .. but I see other non-Rust entries (shellcheck,shfmt,syft) in README.md put `/usr/local/bin` in that column.

I see in `main.sh` that it says
> Copying files to /usr/local/include requires sudo, so do not use it.

taiki-e: `protoc` is special. It needs to copy more files other than `protoc` binary itself.
In the case of no such special requirements, I would prefer to use `/usr/local/bin` for non-Rust tools.

.github/workflows/ci.yml
```diff
@@ -68,21 +68,21 @@ jobs:
           tool: ${{ steps.tool-list.outputs.tool }}
       # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
       - name: Test bash
-        run: just --version; shfmt --version; protoc --version
+        run: just --version; shfmt --version; protoc --version; osv-scanner --version
```
taiki-e, line 71: The CI file does not need to be changed because [main.sh checks if the installation is successful](https://github.com/taiki-e/install-action/blob/d140130aeedb5a946a5769684d32e3a33539f226/main.sh#L669).

https://github.com/taiki-e/install-action/actions/runs/7155101623/job/19483349168#step:5:181

The additional checks in these lines are for compatibility with various shells (e.g., windows shells and unix shells have different path formats), and it is sufficient to check each tool for different installation ways one by one.

jayvdb: Ok, I'll remove these - they were only added due to the request above.

README.md
jayvdb: Is this ok @taiki-e , or should I use `/usr/local/bin` ?

taiki-e (approved):
Thanks!

taiki-e:
Published in 2.22.0.

PR #344, opened 2024-01-22
Add `cargo-sort`

jhpratt:


jhpratt:
Not sure why alpine is failing, as it appears as though it was installed successfully. CentOS and Ubuntu 18.04 are both failing because glibc is too old, I presume. I'm not sure if there's anything that can be done about this?

NobodyXu:
> Not sure why alpine is failing, as it appears as though it was installed successfully. CentOS and Ubuntu 18.04 are both failing because glibc is too old, I presume. I'm not sure if there's anything that can be done about this?

How about swithcing to musl for all targets?

That should fix error on ubuntu 18.04 and CentOS, the error on Alpine is probably because glibc cannot be found.

taiki-e:
> Not sure why alpine is failing, as it appears as though it was installed successfully. CentOS and Ubuntu 18.04 are both failing because glibc is too old, I presume. I'm not sure if there's anything that can be done about this?



If a musl binary is provided, please change to use it. If not, please add the tool name to one of the `*_incompat` lists here.

https://github.com/taiki-e/install-action/blob/5be08121a64dab5197b48a185ea5e0aedb0ccbb5/tools/ci/tool-list.sh#L7-L25

jhpratt:
The binaries included are the only three provided, so I've added an incompatibility as requested. CI is now passing.

NobodyXu (approved):

taiki-e:
Published in 2.26.0. Thanks @jhpratt!

Issue #302, opened 2023-12-10
error: file-permissions-check failed: executable should be in tools/ directory: main.sh


jayvdb:
https://github.com/taiki-e/install-action/actions/runs/7154757157/job/19482642672?pr=299 failed with

```
info: checking file permissions
Error: file-permissions-check failed: executable should be in tools/ directory
=======================================
main.sh
=======================================
```

I can reproduce this locally

taiki-e:
You need to add publish=false, if you add version to manifest. 

Note that both version and publish were removed in the commit that adopted the version-less manifest (https://github.com/taiki-e/install-action/commit/88c52281ad0c02e9c2f84ca76b7c589db5d7a9e7). If you are re-adding a version, please revert that commit rather than manually adding the fields.

PR #383, opened 2024-02-23
Add wasm-bindgen

Ekleog:
Hey! I need to use wasm-bindgen-test-runner in CI, and thought I'd contribute back implementation of it.

Just to say, the `wasm-bindgen-cli` package actually includes other binaries (wasm-bindgen and wasm2es6js). I'm not a user of these packages and I don't know how useful they'd be in CI, so I didn't focus on them, but… it looks like in order to download multiple packages from a single archive, the user would have to download the archive one time for each binary, and we'd have to maintain the json file once for each binary?

So the main thing I'm unsure of is the naming of the tool. Should I have it be `wasm-bindgen-test-runner`, as there currently can be a single executable per tool? Or should we first support multiple executables per tool and then name it `wasm-bindgen-cli`, to match the way it can currently be installed via `binstall`? (I'll say that I probably won't have time to do the second option)

Also I'll take advantage to ask: I also wanted to integrate `sqlx-cli`, but AFAICT [its releases do not have binaries published](https://github.com/launchbadge/sqlx/releases/tag/v0.7.3). Do I understand correctly that install-action is thus unable to help here?


Ekleog:
I think this is ready for review! :D There is just the `tidy` test that is failing, but it is not related to this PR and seems to be due to the way CI works. I tried simply moving `main.sh` to `tools/`, but then other things started failing. Not knowing whether that is actually the way you want to go, I just gave up on fixing CI and left this one red :)

main.sh
```diff
@@ -667,8 +667,8 @@ for tool in "${tools[@]}"; do
     fi
     # cargo-udeps 0.1.30 and wasm-pack 0.12.0 do not support --version option.
     case "${tool}" in
-        biome) rx "${tool_bin}" --version || true ;; # biome up to 1.2.2 exits with 1 on --version and  --help
-        cargo-careful | cargo-machete) ;;            # cargo-careful 0.3.4 and cargo-machete 0.5.0 do not support neither --version nor --help option.
+        biome) rx "${tool_bin}" --version || true ;;                 # biome up to 1.2.2 exits with 1 on --version and  --help
```
NobodyXu, line 670: I think the space added before comment could be the reason why tidy has failed

tools/codegen/Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name = "install-action-internal-codegen"
+version = "0.0.1"
```
NobodyXu, line 3: I don't think this internal crate should have a version, rather it should have:

```suggestion
publish = false
```

NobodyXu (changes requested):

tools/codegen/Cargo.toml
taiki-e: Neither field is needed for version-less manifests (https://github.com/rust-lang/cargo/pull/12786).

main.sh
Ekleog: Actually I'm pretty sure tidy has failed due to this line:
```
Error: file-permissions-check failed: executable should be in tools/ directory
```

I added the spaces when tidy told me to :)

tools/codegen/Cargo.toml
Ekleog: I can remove it if you want, but my local distribution-provided version of cargo failed due to lack of the `version` field.

TBH I don't really care, I just added it en-passant, thinking it could help other people have one fewer thing to think about when contributing :)

main.sh
taiki-e: > ```
> Error: file-permissions-check failed: executable should be in tools/ directory
> ```

This is due to the fact that you have added a version to codegen. See https://github.com/taiki-e/install-action/issues/302.

tools/codegen/Cargo.toml
taiki-e: To be honest, I'm surprised that codegen worked with the old rustc.

We are basically not going to guarantee that codegen will work on older rustc. (If you want to do that you need to check it with CI, otherwise it makes no sense as it can be easily broken.)

Ekleog:
Ooh I see, thank you for your answers!

I just removed the drive-by addition of the `version` field, if as you say codegen isn't supposed to work with older rustc versions and version-less manifests are not going away anyway :)

Now CI passed, please let me know if you think there are other changes to make to this PR!

The main question I have is still how to deal with the multiple-binaries-inside-a-single-crate situation: currently it's possible to install things via `wasm-bindgen-cli` which will proxy through `binstall`, but AFAICT I can't just add a manifest that'd install multiple binaries, so I can't just name the tool `wasm-bindgen-cli` :/

NobodyXu (approved):
LGTM, but I'd let taiki-e to merge this

taiki-e:
LGTM assuming we merge this with a single binary. However, I would like to do a little evaluation to see how difficult it would be to support multiple binaries before merging. (We have already done something more complicated for protoc but that requires some manually written code.)

taiki-e:
Ok, I pushed a commit to support multiple binaries:

```
info: installing wasm-bindgen@latest
info: downloading https://github.com/rustwasm/wasm-bindgen/releases/download/0.2.91/wasm-bindgen-0.2.91-x86_64-unknown-linux-musl.tar.gz
info: verifying sha256 checksum for wasm-bindgen-0.2.91-x86_64-unknown-linux-musl.tar.gz
info: wasm-bindgen-test-runner installed at /home/runner/.cargo/bin/wasm-bindgen-test-runner
info: wasm-bindgen installed at /home/runner/.cargo/bin/wasm-bindgen
info: wasm2es6js installed at /home/runner/.cargo/bin/wasm2es6js
+ wasm-bindgen --version
wasm-bindgen 0.2.91 (fe8bc949b)
```

taiki-e:
Published in 2.28.0. Thanks for the PR!

Ekleog:
Awesome, thank you for all the improvements you did to support this use case!

PR #411, opened 2024-03-17
Add support for git-cliff

tranzystorekk:
Adds support for `git-cliff`, initially discussed in https://github.com/orhun/git-cliff-action/issues/18

I'm not sure if it's a bug or just undocumented behavior, but I couldn't add both gnu and musl targets because the manifest tool removes all gnu targets if they coexist with musl and then errors out later that no gnu targets exits...


taiki-e (approved):
Thanks!

taiki-e:
> I'm not sure if it's a bug or just undocumented behavior, but I couldn't add both gnu and musl targets because the manifest tool removes all gnu targets if they coexist with musl and then errors out later that no gnu targets exits...

This is the intended behavior: if both linux-musl and linux-gnu binaries are present, install-action will always install the linux-musl binary for compatibility with older glibc, so there is no need to leave information about the linux-gnu binaries in the final manifest. If all releases have both binaries, the linux-gnu binary is never used, so you will get an error that it is unused. 

However, it is indeed not documented in the development guide...

taiki-e:
Published in 2.29.0.

PR #422, opened 2024-03-26
Add typos

jayvdb:
https://github.com/crate-ci/typos is installable via binstall via QuickInstall as "typos-cli".

I am not sure what criteria is used for adding binaries directly to install-action .  Justification is this is becoming a very popular tool to use in Rust crates due to very low noise.

Also should this be called "typos", which is the binary name , or stick with the crate name "typos-cli".


taiki-e:
> Also should this be called "typos", which is the binary name , or stick with the crate name "typos-cli".

I feel calling it "typos" instead of *-cli is consistent with crate, which uses similar patterns like wasmtime and wasm-bindgen

taiki-e:
It seems you have opened two PRs on spellchecking tools, do you need both?

jayvdb:
We use both. typos is much better for a first pass, as it recommends correct fixes most of thev time, helping get the right spellings quickly. cargo-spellcheck runs after, then typically picks up the rest of the problems, but its recommendations are not as good, its dicts are incomplete, and whitelisting invovles the more complex myspell dic/aff format. We're also going to add cspell-cli, probably after typos but before cargo-spellcheck. And maybe LanguageTool aa a final later.

taiki-e (approved):
Thanks.

Issue #257, opened 2023-10-19
Request: Add hyperfine


KSXGitHub:
https://github.com/sharkdp/hyperfine

taiki-e:
I would accept a PR to support this.

See [DEVELOPMENT.md](https://github.com/taiki-e/install-action/blob/main/DEVELOPMENT.md) for how to add support for a new tool.

taiki-e:
Added in 2.32.0 (#425).

PR #425, opened 2024-03-27
Add hyperfine

jayvdb:
Closes https://github.com/taiki-e/install-action/issues/257


taiki-e:
I guess you have to add a `bin` field to the base manifest like cargo-audit.
https://github.com/taiki-e/install-action/blob/b208058c0b6867428f1ed35938eec7ed87ced1ce/tools/codegen/base/cargo-audit.json#L6

taiki-e (approved):
Thanks

PR #430, opened 2024-03-31
Add editorconfig-checker

jayvdb:
Add https://github.com/editorconfig-checker/editorconfig-checker

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -23,6 +23,18 @@ See JSON files in `tools/codegen/base` directory for examples of the manifest.
 > GITHUB_TOKEN=$(gh auth status --show-token 2>&1 | sed -n 's/^.*Token: \(.*\)$/\1/p') ./tools/manifest.sh <tool>
 > ```
 
+## Local testing
+
+To test locally after creating the manifest, run
+
+```sh
+INPUT_TOOL=tool_name INPUT_CHECKSUM=false ./main.sh
+```
+
+As this attempt to install into `/usr/local/bin/`, an error as follows suggests that the script is working:
+
+> mv: cannot create regular file '/usr/local/bin/tool_name': Permission denied
+
 ## Troubleshooting
 
 If one of the CI builds fails due to a bin path or release asset_name, fix the problem in the base
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -23,18 +23,6 @@ See JSON files in `tools/codegen/base` directory for examples of the manifest.
 > GITHUB_TOKEN=$(gh auth status --show-token 2>&1 | sed -n 's/^.*Token: \(.*\)$/\1/p') ./tools/manifest.sh <tool>
 > ```
 
-## Local testing
-
-To test locally after creating the manifest, run
-
-```sh
-INPUT_TOOL=tool_name INPUT_CHECKSUM=false ./main.sh
-```
-
-As this attempt to install into `/usr/local/bin/`, an error as follows suggests that the script is working:
-
-> mv: cannot create regular file '/usr/local/bin/tool_name': Permission denied
-
 ## Troubleshooting
 
 If one of the CI builds fails due to a bin path or release asset_name, fix the problem in the base
--- a/main.sh
+++ b/main.sh
@@ -68,9 +68,9 @@ download_and_extract() {
     local tmp
     case "${tool}" in
         # xbuild's binary name is "x", as opposed to the usual crate name
-        xbuild) installed_bin=("${bin_dir}/x") ;;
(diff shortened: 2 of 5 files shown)


main.sh
```diff
@@ -69,6 +69,8 @@ download_and_extract() {
     case "${tool}" in
         # xbuild's binary name is "x", as opposed to the usual crate name
         xbuild) installed_bin=("${bin_dir}/x") ;;
+        # editorconfig-checker's binary name is renamed below
+        editorconfig-checker) installed_bin=("${bin_dir}/${tool}") ;;
```
taiki-e, line 73: It seems editorconfig-checker binary is usually named `ec`? https://github.com/editorconfig-checker/editorconfig-checker?tab=readme-ov-file#what
if so, this should be "${bin_dir}/ec{exe}".

jayvdb: `go install github.com/editorconfig-checker/editorconfig-checker/v2/cmd/editorconfig-checker@latest` installs `editorconfig-checker`

homebrew, nix and arch use `editor-checker` https://repology.org/project/editorconfig-checker/versions

I'm not sure about alpine.  But in general distros are unlikely to let an obscure tool like this have a two letter bin name.

taiki-e: Thanks for the info. Agreed that "editorconfig-checker" is a reasonable binary name here.

alpine: "package name" is "editorconfig-checker", but "binary name" is "ec": https://pkgs.alpinelinux.org/contents?branch=v3.14&name=editorconfig%2dchecker&arch=x86_64&repo=community

arch: "package name" is "editorconfig-checker", but "binary names" are both "ec" and "editor-checker": https://archlinux.org/packages/extra/x86_64/editorconfig-checker/ (click "Package Contents" section to show list)

homebrew: both "package name" and "binary names" are "editorconfig-checker": https://github.com/Homebrew/homebrew-core/blob/675ca659ebe69e0829d59104c2b74089e0acf6f6/Formula/e/editorconfig-checker.rb#L31

nix: 
- NixOS/nixpkgs: IIUC both "package name" and "binary names" are "editorconfig-checker": https://github.com/NixOS/nixpkgs/blob/release-23.11/pkgs/development/tools/misc/editorconfig-checker/default.nix
- editorconfig-checker/editorconfig-checker/default.nix: "package name" is "editorconfig-checker", but "binary names" are both "ec" and "editor-checker" (ec is symlink): https://github.com/editorconfig-checker/editorconfig-checker/blob/0e32afcec97ba925d3c5f6530186159e3f37bcb0/default.nix#L19

taiki-e, line 73:
```suggestion
        xbuild) installed_bin=("${bin_dir}/x${exe}") ;;
        # editorconfig-checker's binary name is renamed below
        editorconfig-checker) installed_bin=("${bin_dir}/${tool}${exe}") ;;
```

tools/codegen/base/editorconfig-checker.json
```diff
@@ -0,0 +1,27 @@
...
+      "bin": "bin/ec-darwin-amd64"
+    },
+    "aarch64_macos": {
+      "asset_name": "ec-darwin-arm64.tar.gz",
+      "bin": "bin/ec-darwin-arm64"
+    },
+    "x86_64_windows": {
+      "asset_name": "ec-windows-amd64${exe}.tar.gz",
+      "bin": "bin/ec-windows-amd64${exe}"
+    }
```
taiki-e, line 25:
```suggestion
    },
    "aarch64_windows": {
      "asset_name": "ec-windows-arm64${exe}.tar.gz",
      "bin": "bin/ec-windows-arm64${exe}"
    }
```

DEVELOPMENT.md
```diff
@@ -23,6 +23,18 @@ See JSON files in `tools/codegen/base` directory for examples of the manifest.
 > GITHUB_TOKEN=$(gh auth status --show-token 2>&1 | sed -n 's/^.*Token: \(.*\)$/\1/p') ./tools/manifest.sh <tool>
 > ```
 
+## Local testing
```
taiki-e, line 26: `main.sh` is a script that is likely to change contants outside the repository, so I don't think it should be run locally, basically, although it would be fine if you are testing it within an isolated environment, such as a container.

Anyway, I don't want to require contributors to run tests locally.

jayvdb: Ok, I've removed those additions

taiki-e (approved):
Thanks!

taiki-e:
Published in 2.32.0.

main.sh
jayvdb: Also is looks like they are going to discontinue using `ec` https://github.com/editorconfig-checker/editorconfig-checker/pull/255

Issue #440, opened 2024-04-04
Update table in readme to mention editorconfig-checker


taiki-e:
I noticed that #430 did not update the table in readme.

cc @jayvdb

jayvdb:
Sorry I forgot that.

I was thinking, it wouldnt be hard to regenerate that table from the Rust, if there were a few more fields added to the base manifest files.  I'd be happy to do that over the weekend if you were interested in that.

taiki-e:
> I was thinking, it wouldnt be hard to regenerate that table from the Rust, if there were a few more fields added to the base manifest files.

This sounds like a great idea!

PR #473, opened 2024-05-05
Generate Markdown table of tools

jayvdb:
Closes https://github.com/taiki-e/install-action/issues/440

This is still a bit of a mess, and the new logic is in the wrong spot - it is currently designed to easily update the existing manifests rather than supporting creating new manifests.

Key question atm is whether it is OK to add new fields `website` & `license_markdown` to both the base info and manifest structs.

Another is whether it is OK to create a `lib.rs` ?  I moved everything to the `lib.rs` , but would be happy to move the unmodified parts back to `main.rs` to reduce the size of the change.


taiki-e:
Thanks!

> Key question atm is whether it is OK to add new fields `website` & `license_markdown` to both the base info and manifest structs.

I'm okay with adding those as those are small enough information.

> Another is whether it is OK to create a `lib.rs` ? I moved everything to the `lib.rs` , but would be happy to move the unmodified parts back to `main.rs` to reduce the size of the change.

I'm okay with either.

jayvdb:
Great.

Next two questions:

1. Is it ok to generate as `TOOLS.md` , and link to it from `README.md`?    IMO the list is already getting long.   I dont mind merging it into the README.md . It is only a bit more work to do that.

2. Should the markdown generator be a separate tool , or run at the end of the existing `main.rs` ?

taiki-e:
> Is it ok to generate as `TOOLS.md` , and link to it from `README.md`?    IMO the list is already getting long.   I dont mind merging it into the README.md . It is only a bit more work to do that.

The current list is indeed long, so splitting it into separate files is okay.

> Should the markdown generator be a separate tool , or run at the end of the existing `main.rs` ?

I think the latter is better at this time because it eliminates the need for changes to the CI script.

jayvdb:
IIUC, the CI loops over each tool, in which case putting the markdown generation in the `main.rs` will mean it does that step many times - once per tool, instead of only once at the end.  Anyways, I'll put it in `main.rs`, but put the logic in `lib.rs`, so it can become a separate bin later.

taiki-e:
> IIUC, the CI loops over each tool, in which case putting the markdown generation in the `main.rs` will mean it does that step many times - once per tool, instead of only once at the end.

Oh, you are right. That is indeed inefficient and would be better to separate them.

TOOLS.md
```diff
@@ -0,0 +1,47 @@
+# Tools
```
taiki-e, line 1: It seems that the new table lost entries that are not managed by manifest from the previous, such as nextest, valgrind.

jayvdb: I've done a comparison now, and it is only nextest and valgrind from README.md that dont appear in TOOLS.md. 

- nextest is easy - we can create a base manifest for it, to get the bins at https://github.com/nextest-rs/nextest/releases/tag/cargo-nextest-0.9.70

- That leaves `valgrind`, which is special as it isnt on GitHub.

Not too hard to have a special case for `valgrind`.

But, I have also noticed that `cargo-watch` & `watchexec-cli` is in https://github.com/taiki-e/install-action/blob/689459d9ffef015a7fbaef7f3b6b9f053f80a64d/tools/ci/tool-list.sh#L135-L138 , but these are not in README.md .  Any others that you know about ?

cargo-watch can also be a base manifest, getting bins from https://github.com/watchexec/cargo-watch/releases/tag/v8.5.2

watchexec-cli might be the bins at https://github.com/watchexec/watchexec/releases/tag/v2.1.1 - need to double check that one.

I see quite a few other system tools that are installed by https://github.com/taiki-e/install-action/blob/main/main.sh .  Should we be mentioning them also?  `jq`, `curl`, `tar` also ?

They would be more special cases, which is fine.

- jq could be a manifest - see https://github.com/jqlang/jq/releases/tag/jq-1.7.1
- https://github.com/curl/curl/releases/tag/curl-8_7_1 doesnt contain binaries
- Maybe we could be replacing `tar` with https://github.com/vadorovsky/packy/releases/tag/v0.0.3 , if it released more arch binaries, and it uploaded untar'd bins to avoid chicken and egg problem.

taiki-e: > But, I have also noticed that `cargo-watch` & `watchexec-cli` is in

tools/ci/tool-list.sh is only for testing. tools/publish.sh is for "Supported tools".

https://github.com/taiki-e/install-action/blob/689459d9ffef015a7fbaef7f3b6b9f053f80a64d/tools/publish.sh#L105-L106

> I see quite a few other system tools that are installed by https://github.com/taiki-e/install-action/blob/main/main.sh . Should we be mentioning them also? jq, curl, tar also ?

Let's ignore them at this PR, as they are for incomplete runners and are not usually installed by us.

tools/codegen/src/tools-markdown.rs
```diff
@@ -0,0 +1,215 @@
...
+            installed_to: InstalledTo::Cargo,
+            installed_from: InstalledFrom::Binstall,
+            platforms: Platforms::all(),
+            repository: "https://github.com/nextest-rs/nextest".to_string(),
+            license_markdown: "[Apache-2.0](https://github.com/nextest-rs/nextest/blob/HEAD/LICENSE-APACHE) OR [MIT](https://github.com/nextest-rs/nextest/blob/HEAD/LICENSE-MIT)".to_string()
+        },
+        ReadmeEntry {
+            name: "valgrind".to_string(),
+            alias: None,
+            website: "https://nexte.st/".to_string(),
```
jayvdb, line 45: to be fixed

tools/codegen/src/tools-markdown.rs
```diff
@@ -0,0 +1,215 @@
...
+            installed_to: InstalledTo::Snap,
+            installed_from: InstalledFrom::Snap,
+            platforms: Platforms {
+                linux: true,
+                ..Default::default()
+            },
+            repository: "https://github.com/nextest-rs/nextest".to_string(),
+            license_markdown: "[Apache-2.0](https://github.com/nextest-rs/nextest/blob/HEAD/LICENSE-APACHE) OR [MIT](https://github.com/nextest-rs/nextest/blob/HEAD/LICENSE-MIT)".to_string()
```
jayvdb, line 53: to be fixed

tools/codegen/src/tools-markdown.rs
```diff
@@ -0,0 +1,215 @@
...
+        };
+        tools.push(readme_entry);
+    }
+
+    println!("# Tools");
+    println!();
+    println!("| Name | Where binaries will be installed | Where will it be installed from | Supported platform | License |");
+    println!("| ---- | -------------------------------- | ------------------------------- | ------------------ | ------- |");
+
+    tools.sort();
```
jayvdb, line 109: try using a sort_by to improve performance and be able to delete some derives

tools/codegen/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Version {
+    pub major: Option<u64>,
+    pub minor: Option<u64>,
+    pub patch: Option<u64>,
+    pub pre: semver::Prerelease,
+    pub build: semver::BuildMetadata,
+}
```
jayvdb, line 34: fwiw, `git diff --color-moved` is a good way to check changes to the items which were moved to lib.rs.

Or if you like, I can create a separate PR to move them to lib.rs , but it still involves a lot of changes because `pub` needed to be added to the struct members

tools/codegen/src/main.rs
```diff
@@ -82,17 +84,37 @@ fn main() -> Result<()> {
...
+            .into_json::<crates_io::Crate>()?;
+        let latest_version = &info.versions[0].num;
+        crates_io_version_detail = Some(
+            download(&format!("https://crates.io/api/v1/crates/{crate_name}/{latest_version}"))?
+                .into_json::<crates_io::VersionMetadata>()?
+                .version,
         );
+
+        if let Some(crate_repository) = info.crate_.repository.clone() {
+            // cargo-dinghy being fixed at https://github.com/sonos/dinghy/pull/231
```
jayvdb, line 105: update the wording to reflect it is now merged

tools/codegen/src/main.rs
```diff
@@ -82,17 +84,37 @@ fn main() -> Result<()> {
...
+            if crate_name != "cargo-dinghy"
+                && !crate_repository
+                    .to_lowercase()
+                    .starts_with(&base_info.repository.to_lowercase())
+            {
+                panic!("warning: crate metadata repository {crate_repository} differs from base manifest");
+            }
+        } else if crate_name != "zola" {
+            panic!("warning: crate metadata does not include a repository");
+        }
```
jayvdb, line 115: IMO we should remove the crate name from zola in the base manifest, as it is only a dummy entry with no metadata https://crates.io/crates/zola

tools/codegen/src/main.rs
```diff
@@ -82,17 +84,37 @@ fn main() -> Result<()> {
...
+
+        if let Some(crate_repository) = info.crate_.repository.clone() {
+            // cargo-dinghy being fixed at https://github.com/sonos/dinghy/pull/231
+            if crate_name != "cargo-dinghy"
+                && !crate_repository
+                    .to_lowercase()
+                    .starts_with(&base_info.repository.to_lowercase())
+            {
+                panic!("warning: crate metadata repository {crate_repository} differs from base manifest");
+            }
```
jayvdb, line 112: need to remove "warning: " since it is panicing

The rationale for panic-ing in this case is https://github.com/taiki-e/install-action/issues/488 .  If the repo has moved, that should require human intervention.

tools/codegen/src/main.rs
```diff
@@ -130,6 +152,66 @@ fn main() -> Result<()> {
...
+        manifests.license_markdown = license_markdown;
+    } else if let Some(detail) = crates_io_version_detail {
+        if let Some(license) = detail.license {
+            if let Some(license_markdown) =
+                get_license_markdown(&license, &repo.to_string(), &repo_info.default_branch)
+            {
+                manifests.license_markdown = license_markdown;
+            }
+        } else {
+            panic!("No license found in crate metadata; set license_markdown manually")
```
jayvdb, line 203: revise slightly to indicate: license markdown couldn't be created from crate metadata

tools/codegen/src/main.rs
```diff
@@ -130,6 +152,66 @@ fn main() -> Result<()> {
...
+        }
+    } else if let Some(license) = repo_info.license {
```
jayvdb, line 205: oops; in the rewrite I dropped a final check at the bottom of this to panic if license_markdown is never set

tools/codegen/src/main.rs
```diff
@@ -561,6 +636,35 @@ fn download_github(url: &str) -> Result<ureq::Response> {
     Err(last_error.unwrap().into())
 }
 
+#[allow(clippy::missing_panics_doc)]
+pub fn github_head(url: &str) -> Result<()> {
+    eprintln!("fetching head of {url} ..");
```
jayvdb, line 641: this & the eprintln for failure is a bit noisy, especially as the license algorithm expects this to fail quite often as it tries to detect where to find the license file.

TOOLS.md
```diff
@@ -0,0 +1,49 @@
...
+| [**cargo-careful**](https://github.com/RalfJung/cargo-careful) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/RalfJung/cargo-careful/releases) | Linux, macOS, Windows | [MIT](https://github.com/RalfJung/cargo-careful/blob/master/LICENSE-MIT) OR [Apache-2.0](https://github.com/RalfJung/cargo-careful/blob/master/LICENSE-APACHE) |
+| [**cargo-cyclonedx**](https://cyclonedx.org/) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/CycloneDX/cyclonedx-rust-cargo/releases) | Linux, macOS, Windows | [Apache-2.0](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/main/LICENSE) |
+| [**cargo-deny**](http://embark.rs) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/EmbarkStudios/cargo-deny/releases) | Linux, macOS, Windows | [MIT](https://github.com/EmbarkStudios/cargo-deny/blob/main/LICENSE-MIT) OR [Apache-2.0](https://github.com/EmbarkStudios/cargo-deny/blob/main/LICENSE-APACHE) |
```
taiki-e, line 10: Could you make it link to the repository instead of the website as before?

For cargo-deny, it would became annoying: auto-redirected and then have to scroll down the page to find the cargo-deny and click on it (finally jump to the cargo-deny repository). For cargo-cycloned, even the mention to the cargo-cycloned is not on the top page of cyclonedx.org.

jayvdb: Is it ok if we do this specifically for cargo-deny and  cargo-cycloned ?

Or, what about I fetch the website and see if the crate name is mentioned, and if not fallback to the repo ?

btw, I am also happy to always use the repo - IMO most people find that most useful that a 'website'.  But the existing readme has website links for biome and others, which I why I tried to detect this

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
taiki-e, line 1: I would prefer not to include this file in the repository, as it may conflict with the contributor's local settings.

Instead is it possible to [use a marker to omit from the toc](https://github.com/taiki-e/cargo-llvm-cov/blob/8a3553b86551eabf9c30c060b1f72a5bbccb98c6/README.md?plain=1#L601)?

TOOLS.md
taiki-e: I think it would be a better idea to make the default a repository link and allow the base manifest to specify a website link for biome and others.

(websites are sometimes created for the time being and have little or no information, or are abandoned.)

Another case where I felt the automatic detection of website links was not sufficient is cargo-audit: the current one is the link to the cargo-audit's readme, and the top page of repository and website have links to various tools, including links to the cargo-audit's readme, so it would be better to specify the current one than the repository link or the website link.

taiki-e: (Btw, AFAIK biome, cargo-audit, shellcheck, and valgrind are the only ones currently not using repository links in main branch.)

jayvdb: Remove the website auto-detection

TOOLS.md
```diff
@@ -0,0 +1,49 @@
...
+| [**valgrind**](https://valgrind.org/) | `/snap/bin` | [snap](https://snapcraft.io/install/valgrind/ubuntu) | Linux | [GPL-2.0](https://sourceware.org/git/?p=valgrind.git;a=blob;f=COPYING;hb=HEAD) |
+| [**wasm-bindgen**](https://github.com/rustwasm/wasm-bindgen) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/rustwasm/wasm-bindgen/releases) | Linux, macOS, Windows | [MIT](https://github.com/rustwasm/wasm-bindgen/blob/main/LICENSE-MIT) OR [Apache-2.0](https://github.com/rustwasm/wasm-bindgen/blob/main/LICENSE-APACHE) |
+| [**wasm-pack**](https://github.com/rustwasm/wasm-pack) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/rustwasm/wasm-pack/releases) | Linux, macOS, Windows | [MIT](https://github.com/rustwasm/wasm-pack/blob/master/LICENSE-MIT) OR [Apache-2.0](https://github.com/rustwasm/wasm-pack/blob/master/LICENSE-APACHE) |
+| [**wasmtime**](https://github.com/bytecodealliance/wasmtime) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/bytecodealliance/wasmtime/releases) | Linux, macOS, Windows | [Apache-2.0](https://github.com/bytecodealliance/wasmtime/blob/main/LICENSE) |
```
taiki-e, line 47: Hmm, wasmtime's license is "Apache-2.0 WITH LLVM-exception" to be exact, but the auto-detection doesn't seem to recognize it.

jayvdb: Let me double check that - I suspect there is a field in the spdx data which contains exceptions.

jayvdb: Fixed

taiki-e: The fix does not seem to have been applied to this file yet.

TOOLS.md
```diff
@@ -0,0 +1,49 @@
...
+| [**cargo-rdme**](https://github.com/orium/cargo-rdme) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/orium/cargo-rdme/releases) | Linux, macOS, Windows | [MPL-2.0](https://github.com/orium/cargo-rdme/blob/main/LICENSE.md) |
+| [**cargo-sort**](https://github.com/DevinR528/cargo-sort) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/DevinR528/cargo-sort/releases) | Linux, macOS, Windows | [Apache-2.0 OR MIT](https://github.com/DevinR528/cargo-sort/blob/55ec89082466f6bb246d870a8d56d166a8e1f08b/Cargo.toml#L5) |
+| [**cargo-spellcheck**](https://github.com/drahnr/cargo-spellcheck) | `$CARGO_HOME/bin` | [GitHub Releases](https://github.com/drahnr/cargo-spellcheck/releases) | Linux, Windows | [MIT](https://github.com/drahnr/cargo-spellcheck/blob/master/LICENSE-MIT) OR [Apache-2.0](https://github.com/drahnr/cargo-spellcheck/blob/master/LICENSE-APACHE) |
```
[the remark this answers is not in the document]
jayvdb: ya, for some reason they dont mention that in the crate metadata.
https://crates.io/crates/cargo-spellcheck says only `[MIT ](https://choosealicense.com/licenses/mit)OR [Apache-2.0](https://choosealicense.com/licenses/apache-2.0)`
I'll override this one.

TOOLS.md
```diff
@@ -0,0 +1,49 @@
+# Tools
+
+| Name | Where binaries will be installed | Where will it be installed from | Supported platform | License |
+| ---- | -------------------------------- | ------------------------------- | ------------------ | ------- |
+| [**biome**](https://biomejs.dev) | `/usr/local/bin` | [GitHub Releases](https://github.com/biomejs/biome/releases) | Linux, macOS, Windows | [Apache-2.0](https://github.com/biomejs/biome/blob/main/LICENSE-APACHE) |
```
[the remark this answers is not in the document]
jayvdb: The github metadata only includes one license, in this case Apache, and it looks like there isnt any indicator that github detected more licenses.

`RepoLicense { key: Some("apache-2.0"), name: Some("Apache License 2.0"), node_id: Some("MDc6TGljZW5zZTI="), spdx_id: Some("Apache-2.0") }`.

I'll add an override in for this.

taiki-e (approved):
Looks great! Thanks @jayvdb!

PR #484, opened 2024-05-19
Avoid using GITHUB_TOKEN with crates.io

jayvdb:


tools/codegen/src/main.rs
```diff
@@ -527,7 +527,9 @@ fn replace_vars(
     Ok(s)
 }
 
-fn download(url: &str) -> Result<ureq::Response> {
+/// Download using GITHUB_TOKEN.
+#[allow(clippy::missing_panics_doc)]
```
jayvdb, line 531: I feel these are ok, as the panic will not occur, so it is silly to document that it can panic.

taiki-e (approved):

PR #485, opened 2024-05-19
Allow skipping existing versions

jayvdb:
As part of development of https://github.com/taiki-e/install-action/pull/473 it was necessary to process all tools, and I didnt want to fetch all of the binaries.  This feature makes that possible.
I am happy to add docs about it wherever would be sensible.


taiki-e (approved):
Thanks!

PR #508, opened 2024-06-05
Add cargo-zigbuild

jayvdb:
https://github.com/rust-cross/cargo-zigbuild


NobodyXu (changes requested):
You would also need to `pip install ziglang` before `cargo-zigbuild` is actually usable.

jayvdb:
well the assumption is that https://github.com/goto-bus-stop/setup-zig or similar has been also run, either before or after.

NobodyXu:
> well the assumption is that https://github.com/goto-bus-stop/setup-zig or similar has been also run, either before or after.

If I am the user, I would assume it has everything installed for cargo-zigbuild to be used.

You can special case cargo-zigbuild by doing a `pip install ziglang`

jayvdb:
My understanding of this action is that it does not do extra installs that might be required to make a tool work.
e.g. installing `cargo-valgrind` does not also install `valgrind` ; installing `xbuild` does not also install `adb` and other bits of llvm not present on the github runner.

It is up to the user of this action to install dependencies that are missing, and that way they can choose the version they want of those dependencies.

I would support adding an extra tool called "ziglang", which would be a special case also like "valgrind".  We could support installing different versions from https://pypi.org/project/ziglang/#history - that looks like a reasonable approach, but Python wheels are a messy format for this given that `zig` is a single binary, and pip is slow.  Importing the binary URLs from https://ziglang.org/download/index.json would be more in tune with the way this action works, allowing runtime verification of the checksums.

NobodyXu (approved):
Yeah that makes sense, having a special support for ziglang along with cargo-zigbuild

jayvdb:
Thanks @NobodyXu .

taiki-e:
Published in 2.36.0. Thanks @jayvdb for the PR and @NobodyXu for the review!

PR #512, opened 2024-06-08
Add xh

jayvdb:
https://github.com/ducaale/xh , curl clone

### DIFF
--- a/.github/.cspell/project-dictionary.txt
+++ b/.github/.cspell/project-dictionary.txt
@@ -27,5 +27,6 @@ udeps
 wasmtime
 watchexec
 xbuild
+xh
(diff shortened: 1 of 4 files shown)


NobodyXu (commented):
Do we really want a curl clone in this action?
Don't we already have curl?

NobodyXu:
I have a similar complaint https://github.com/taiki-e/install-action/pull/513#pullrequestreview-2105778288 , really don't think we should support everything here.

NobodyXu (approved):
Thanks!

xh seems quite useful, while curl is enough for many scripts, xh is quite useful for testing restful APIs and a good alternative to httpie

PR #513, opened 2024-06-08
Add jaq

jayvdb:
https://github.com/01mf02/jaq , jq clone


NobodyXu (commented):
Do we really need a jq clone support in this action?

I get that it is an alternative in Rust and I support it, but given that GHA already have jq and it is enough for most use cases, I'm not sure if this is needed since this action fallbacks to cargo-binstall for rust crates.

Sure it is a bit slower than supporting it here directly but most of the time it is fine and good enough, I don't think we should support everything here.

NobodyXu (approved):
I read through jaq description, it seems that they aim to be more correct (regarding floating point, etc), plus it's written in rust and use serde_json to parse json which provides better memory safety than C parser.

It makes sense for medical software and other area which pay great attention to safety and correctness.

NobodyXu:
Thanks for explanation, that makes sense.

It seems that the CI has hit rate limit of GitHub, so it's quite slow.

NobodyXu:
I have rebased the PR against main to fix the CI due to outdated cargo-binstall

PR #517, opened 2024-06-08
Add fallback input option to ensure that fallback is not used

taiki-e:
Add an option to prevent accidentally installing tools that use fallback for installation.

```yaml
- uses: taiki-e/install-action@v2
  with:
    tool: <tool>
    # Possible values:
    # - none: disable all fallback
    # - cargo-binstall (default): cargo-binstall (includes quickinstall)
    fallback: none
```

ss

We could add an option to use cargo-binstall without quickinstall in the future if someone actually wants to use that.

Related discussion: https://github.com/taiki-e/install-action/pull/514

cc @jayvdb

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -10,6 +10,10 @@ inputs:
     description: Whether to enable checksums
     required: false
     default: 'true'
+  fallback:
+    description: Whether to use fallback (none or cargo-binstall)
+    required: false
+    default: 'cargo-binstall'
 
 # Note:
 # - inputs.* should be manually mapped to INPUT_* due to https://github.com/actions/runner/issues/665
@@ -22,3 +26,4 @@ runs:
       env:
         INPUT_TOOL: ${{ inputs.tool }}
         INPUT_CHECKSUM: ${{ inputs.checksum }}
+        INPUT_FALLBACK: ${{ inputs.fallback }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
         run: just --version & shfmt --version & protoc --version
         shell: cmd
         if: startsWith(matrix.os, 'windows')
-      # We use the version output to check the version of binstall, but they
+      # We use the version output to check the version of cargo-binstall, but they
       # several times change the version output format in the past so we need to
       # check it with CI. (e.g., 0.14.0->0.16.0 update change it
       # from "cargo-binstall <version>" to "<version>")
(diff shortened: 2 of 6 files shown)


action.yml
```diff
@@ -22,3 +26,4 @@ runs:
       env:
         INPUT_TOOL: ${{ inputs.tool }}
         INPUT_CHECKSUM: ${{ inputs.checksum }}
+        INPUT_FALLBACK: ${{ inputs.fallback }}
```
NobodyXu, line 29: I'd suggest `INPUT_STRATEGIES` and pass it directly to `cargo-binstall`.

User can specify `crate-meta-data` for trying to download it from official maintainer, `quick-install` from third-party quickinstall, `compile` for from source.

If empty, then disable cargo-binstall

https://github.com/cargo-bins/cargo-binstall/blob/48ee0b0e3e646f7f0cfb2428f46dbcd32afe979b/crates/bin/src/args.rs#L418

jayvdb: getting approval for cargo-binstall would be worth attempting IMO if its cargo-quickinstall was disabled.
i.e. trying to explain to my cyber-security team that we're using install-action which falls back to cargo-binstall which falls back to cargo-quickinstall is too hard, and it is that cargo-quickinstall stage which becomes "oh its just grabbing some binary from somewhere on the internet, but they are good people running it, trust them".  The fall back to cargo-binstall is explainable if it is "it builds from source 100%"

NobodyXu: That's good to hear, disabling quickinstall is pretty easy, you just need to pass `--strategies crate-meta-data,compile` and it will only use the prebuilt from the official repository specified in `Cargo.toml` on https://crates.io , or fallback to `cargo-install`.

`--strategies compile` would then be equivalent to launching `cargo-install`

taiki-e: IMO, for a reviewer who is not familiar with the details of cargo-binstall's API but needs to review code that uses this action, something like `strategies: ''`/`strategies: crate-meta-data,compile` is not at all clear at first glance.

NobodyXu: That's true, maybe we can pass `--disable-strategies` instead.

`--disable-strategies quick-install` is equivalent to `--strategies crate-meta-data,compile`

jayvdb (approved):

taiki-e:
Considering there is a proposal to add a non-cargo-binstall fallback (https://github.com/taiki-e/install-action/issues/526) too, I think this option is good as is. We can discuss how to pass additional flags to cargo-binstall in another issue.

PR #518, opened 2024-06-08
Add manifest for cargo-nextest

### LINKED ISSUE #487: Creating a manifest for nextest
nextest is currently a special case found at https://github.com/taiki-e/install-action/pull/473#discussion_r1606066064

It seems reasonable for this popular tool to be supported directly by this action, so binstall doesnt need to be used for it.

Is there anything that needs to be done in order to avoid breaking existing users who are using the "nextest" tool - i.e. switching them from binstall to native manifest installs?

taiki-e:
Closes #487

This is basically a revert of https://github.com/taiki-e/install-action/pull/183 and a fork of https://github.com/taiki-e/install-action/pull/182 (changed to apply only for nextest, see my comment in it for the reason).
This improves security, performance, robustness of installation.

cc @sunshowers @jayvdb


main.sh
```diff
@@ -212,6 +234,21 @@ read_manifest() {
...
+            elif [[ "${host_env}" == "gnu" ]]; then
+                # TODO: don't hardcode tool name and use 'prefer_linux_gnu' field in base manifest.
+                case "${tool}" in
+                    cargo-nextest | nextest)
+                        # TODO: don't hardcode required glibc version
+                        required_glibc_version=2.27
+                        higher_glibc_version=$(sort <<<"${required_glibc_version}"$'\n'"${host_glibc_version}" -Vu | tail -1)
+                        if [[ "${higher_glibc_version}" == "${host_glibc_version}" ]]; then
+                            # musl build of nextest is slow, so use glibc build if host_env is gnu.
+                            # https://github.com/taiki-e/install-action/issues/13
```
taiki-e, line 246: @sunshowers It has been over a year since the update to musl 1.2 in rustc, is my understanding correct that this performance issue still exists?

jayvdb (approved):

seanmonstar:
I don't know if it's related, but the timing lines up: as of 2 days ago, reqwest CI started failing on i686-pc-windows-gnu, and 86_64. It seems to fail to install the correct version of nextest: https://github.com/seanmonstar/reqwest/actions/runs/9438070294/job/26022148081#step:9:22

taiki-e:
A real problem is that msys64's bash is used here, which seems to introduce a lot of differences from the default one, especially that HOME will be a different one on the string.


https://github.com/seanmonstar/reqwest/actions/runs/9438070294/job/26022148081#step:4:2
```
  echo "C:\msys64\mingw32\bin" >> $GITHUB_PATH
  echo "C:\msys64\usr\bin" >> $GITHUB_PATH
```


Default one is: https://github.com/taiki-e/install-action/actions/runs/9465237915/job/26074336422#step:5:18

```
  shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
```

`$HOME` on bash is  `/c/Users/runneradmin`.


When msys64 is used: https://github.com/seanmonstar/reqwest/actions/runs/9438070294/job/26022148081#step:8:17

```
  shell: C:\msys64\usr\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
```

`$HOME` on bash is  `/home/runneradmin`.

And the error you saw is due to the fact that a path starting `/home/runneradmin/*` that added to PATH by using the GitHub-provided mechanism for adding PATH (GITHUB_PATH) is not recognized by the msys64 bash. (a path starting `/c/Users/runneradmin/*` is recognized)

This will be fixed in https://github.com/taiki-e/install-action/pull/533.

Issue #521, opened 2024-06-08
archlinux test container is broken: jq now requires GLIBC_2.39


jayvdb:
```
Install packages required for installation (jq, curl, and/or tar)
  :: Synchronizing package databases...
   core downloading...
   extra downloading...
  resolving dependencies...
  looking for conflicting packages...
  
  Package (2)      New Version  Net Change  Download Size
  
  extra/oniguruma  6.9.9-1        0.87 MiB       0.21 MiB
  extra/jq         1.7.1-2        0.73 MiB       0.28 MiB
  
  Total Download Size:   0.50 MiB
  Total Installed Size:  1.60 MiB
  
  :: Proceed with installation? [Y/n] 
  :: Retrieving packages...
   jq-1.7.1-2-x86_64 downloading...
   oniguruma-6.9.9-1-x86_64 downloading...
  checking keyring...
  checking package integrity...
  loading package files...
  checking for file conflicts...
  :: Processing package changes...
  installing oniguruma...
  installing jq...
  :: Running post-transaction hooks...
  (1/1) Arming ConditionNeedsUpdate...
jq: /usr/lib/libm.so.6: version `GLIBC_2.39' not found (required by jq)
```

jayvdb:
there was a recent change upstream to jq package https://gitlab.archlinux.org/archlinux/packaging/packages/jq/-/commit/0620080827b37757a876144b701ca7bc48aebbc7

But that doesnt appear to be the problem.
And when I attempt to reproduce locally, it works.
```
> docker run -it alpine
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
d25f557d7f31: Pull complete
Digest: sha256:77726ef6b57ddf65bb551896826ec38bc3e53f75cdde31354fbffb4f25238ebd
Status: Downloaded newer image for alpine:latest
/ # apk --no-cache add jq
fetch https://dl-cdn.alpinelinux.org/alpine/v3.20/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.20/community/x86_64/APKINDEX.tar.gz
(1/2) Installing oniguruma (6.9.9-r0)
(2/2) Installing jq (1.7.1-r0)
Executing busybox-1.36.1-r28.trigger
OK: 9 MiB in 16 packages
/ # jq --help
jq - commandline JSON processor [version 1.7.1]
...
```

NobodyXu:
[...]

PR #525, opened 2024-06-09
Workaround archlinux jq failure

jayvdb:
Fixes https://github.com/taiki-e/install-action/issues/521

### DIFF
--- a/manifests/dprint.json
+++ b/manifests/dprint.json
@@ -25,10 +25,27 @@
   },
   "license_markdown": "[MIT](https://github.com/dprint/dprint/blob/main/LICENSE)",
   "latest": {
-    "version": "0.46.1"
+    "version": "0.46.2"
   },
   "0.46": {
-    "version": "0.46.1"
+    "version": "0.46.2"
+  },
+  "0.46.2": {
+    "x86_64_linux_musl": {
+      "checksum": "bbe9fe8eae9abdcfccdeca97fd8c524efd6137de702ee96e82b0ecb4ad432ebf"
+    },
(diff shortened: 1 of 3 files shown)


jayvdb:
@taiki-e @NobodyXu this is green now

manifests/dprint.json
```diff
@@ -25,10 +25,27 @@
   },
   "license_markdown": "[MIT](https://github.com/dprint/dprint/blob/main/LICENSE)",
   "latest": {
-    "version": "0.46.1"
+    "version": "0.46.2"
```
NobodyXu, line 28: The update of dprint should be in a separate PR, IMO we should wait for the GHA to auto-create and merge it, then remove this commit.

NobodyXu (changes requested):

jayvdb:
That is a catch-22 - the scheduled CI failed because of this https://github.com/taiki-e/install-action/actions/runs/9433528539

NobodyXu:
So we need this PR to merge for the CI to run and auto-merge.

I see, this PR LGTM.

NobodyXu:
There's still a problem of if it will show up in the commit history.

@jayvdb Could you clean up the commit history a bit?

I think doing a rebase merge might make more sense, to preserve the commit history.

cc @taiki-e What do you think?

jayvdb:
I've updated the commit message of the middle commit.
The other two commits are different things, hence they are separate commits.

jayvdb:
Green again

taiki-e (approved):
Thanks!

PR #533, opened 2024-06-11
Workaround msys64 bash's buggy behavior

taiki-e:
Fixes https://github.com/taiki-e/install-action/pull/518#issuecomment-2160736760


seanmonstar:
Thanks so much! <3

taiki-e:
@seanmonstar This fix has been released in 2.38.3. Thanks for the report and sorry for the breakage!

Issue #540, opened 2024-06-15
Speedup CI manifest


NobodyXu:
CI manifest is quite slow, sometimes taking up to 20m.

I think using a dedicated github token might help, `secrets.GITHUB_TOKEN` only provides 1000 API hits per hour, where one created by user gets 5000.

Switching to GraphQL might also help, since we can fetch all the artifacts information when fetching the release information.

NobodyXu:
A more efficient approach would be webhooks, though it can only be created by the repository owner.

We can ask them to help with this project, it would be the most efficient way without having to go through GitHub API, and most GitHub friendly.

The webhook information can be cached on a dedicated github branch, so that the manifest CI won't refetch them.

taiki-e:
cc @jayvdb 

This is especially slow since the recent codegen changes (https://github.com/taiki-e/install-action/pull/473).

> fetch all the artifacts information when fetching the release information.

Currently, we always call requests that have a limit of 100 releases per request in several times to get all the releases. However, if codegen is called with the "latest" argument, as in CI, it is usually sufficient to look at the latest 100 releases.

Eventually we may need to change the overall logic, I guess there are a number of areas that could be made better without such a big deal.

NobodyXu:
> Eventually we may need to change the overall logic

@taiki-e I think using web hook for certain projects and cache them in a dedicated branch would be much efficient.

cc @sunshowers

taiki-e:
By the way, in my understanding, one of the actual causes of the "slowness" is that even though the token has reached the rate limit, that fact is not remembered on the codegen side, and it attempts to download with the token that has reached the rate limit each time the download is invoked, repeatedly.

NobodyXu:
> that fact is not remembered on the codegen side, and it attempts to download with the token that has reached the rate limit each time the download is invoked, repeatedly.

The [binstalk-downloader](https://crates.io/crates/binstalk-downloader) crate checks for rate limit

https://github.com/cargo-bins/cargo-binstall/blob/0d7080e6a9e1ca43a37026b52683e367c416c87f/crates/binstalk-downloader/src/remote.rs#L182

and applies a backoff strategy

https://github.com/cargo-bins/cargo-binstall/blob/0d7080e6a9e1ca43a37026b52683e367c416c87f/crates/binstalk-downloader/src/remote/delay_request.rs#L70

NobodyXu:
[...]

PR #547, opened 2024-06-19
Use asset etag to avoid fetches

jayvdb:
c.f. https://github.com/taiki-e/install-action/issues/540#issuecomment-2170998595

This doesnt reduce the number of API hits - it only halts downloads if the current etag is the same as in the head of the asset fetch.

As a result the manifest CI job goes from 4m30s to ~2m.

Using `If-None-Match` header would also work, but 

1. that would add complexity to `download(..)`, and the logic around the invocation to optionally provide an etag if it was found in the existing manifest.
2. I found using `head` doesnt work, as the Server responded to `ureq::head` with 401 if `If-None-Match` is used; however responds to `wget` head with 304.

IMO it would be worthwhile adding complexity to `download(..)` if more of the fetches also become etag capable, which I believe is possible and worthwhile, but IMO not as valuable as the asset fetches because they are so numerous and voluminous to be likely to prevent features being developed by new contributors who will balk at downloading so much data just to check a new feature against the full set of manifests.

The other major addition is writing the manifest during processing of many versions - some of the larger manifests were failing when I was downloading all of the old assets, and needed to be restarted, and I was loosing too much data/time if it had to restart at the beginning with all the etags missing.

### DIFF
--- a/tools/codegen/src/main.rs
+++ b/tools/codegen/src/main.rs
@@ -30,6 +30,7 @@ fn main() -> Result<()> {
     }
     let package = &args[0];
     let version_req = args.get(1);
+    let version_req_given = version_req.is_some();
     let skip_existing_manifest_versions = std::env::var("SKIP_EXISTING_MANIFEST_VERSIONS").is_ok();
 
     let workspace_root = &workspace_root();
@@ -106,12 +107,7 @@ fn main() -> Result<()> {
         );
 
         if let Some(crate_repository) = info.crate_.repository.clone() {
-            // cargo-dinghy is fixed at https://github.com/sonos/dinghy/pull/231, but not yet released
-            if crate_name != "cargo-dinghy"
-                && !crate_repository
-                    .to_lowercase()
-                    .starts_with(&base_info.repository.to_lowercase())
-            {
+            if !crate_repository.to_lowercase().starts_with(&base_info.repository.to_lowercase()) {
                 panic!("repository {crate_repository} from crates.io differs from base manifest");
             }
         } else if crate_name != "zola" {
@@ -246,7 +242,9 @@ fn main() -> Result<()> {
 
         let reverse_semver = Reverse(semver_version.clone().into());
 
-        if skip_existing_manifest_versions && manifests.map.contains_key(&reverse_semver) {
+        let existing_manifest = manifests.map.get(&reverse_semver).cloned();
+
+        if skip_existing_manifest_versions && existing_manifest.is_some() {
             eprintln!("Skipping {semver_version} already in manifest");
             continue;
         };
@@ -287,19 +285,40 @@ fn main() -> Result<()> {
                 }
             };
 
-            eprint!("downloading {url} for checksum ... ");
+            eprint!("downloading {url} for etag and checksum ... ");
             let download_cache = &download_cache_dir.join(format!(
                 "{version}-{platform:?}-{}",
                 Path::new(&url).file_name().unwrap().to_str().unwrap()
             ));
+            let response = download(&url)?;
+            let etag = response
+                .header("etag")
+                .expect("binary should have an etag")
+                .to_string()
+                .replace('\"', "");
+
+            if let Some(ManifestRef::Real(ref manifest)) = existing_manifest {
+                if let Some(entry) = manifest.download_info.get(&platform) {
+                    if entry.etag == etag {
+                        eprintln!("existing etag matched");
+                        download_info.insert(platform, entry.clone());
+                        continue;
+                    }
+                    eprintln!("warn: existing etag no longer valid.");
+                } else {
+                    eprintln!("existing manifest for {version} is missing platform {platform:?}");
+                }
+            }
+
             if download_cache.is_file() {
                 eprintln!("already downloaded");
                 fs::File::open(download_cache)?.read_to_end(&mut buf)?;
             } else {
-                download(&url)?.into_reader().read_to_end(&mut buf)?;
+                response.into_reader().read_to_end(&mut buf)?;
                 eprintln!("download complete");
                 fs::write(download_cache, &buf)?;
             }
+
             eprintln!("getting sha256 hash for {url}");
             let hash = Sha256::digest(&buf);
             let hash = format!("{hash:x}");
@@ -385,6 +404,7 @@ fn main() -> Result<()> {
 
             download_info.insert(platform, ManifestDownloadInfo {
                 url: Some(url),
+                etag,
                 checksum: hash,
                 bin: base_download_info.bin.as_ref().or(base_info.bin.as_ref()).map(|s| {
                     s.map(|s| {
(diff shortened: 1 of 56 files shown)


tools/codegen/src/main.rs
```diff
@@ -287,19 +285,40 @@ fn main() -> Result<()> {
...
             };
 
-            eprint!("downloading {url} for checksum ... ");
+            eprint!("downloading {url} for etag and checksum ... ");
             let download_cache = &download_cache_dir.join(format!(
                 "{version}-{platform:?}-{}",
                 Path::new(&url).file_name().unwrap().to_str().unwrap()
             ));
+            let response = download(&url)?;
+            let etag = response
```
NobodyXu, line 294: According to [mdn doc for etags](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag), it seems that the etag could contain a `W/` to indicate weak validator is used.

I think we should strip it from the etag stored in manifest.

Example etag:

```
ETag: "33a64df551425fcc55e4d42a148795d9f25f89d4"
ETag: W/"0815"
```

tools/codegen/src/main.rs
```diff
@@ -287,19 +285,40 @@ fn main() -> Result<()> {
...
+                .header("etag")
+                .expect("binary should have an etag")
+                .to_string()
+                .replace('\"', "");
+
+            if let Some(ManifestRef::Real(ref manifest)) = existing_manifest {
+                if let Some(entry) = manifest.download_info.get(&platform) {
+                    if entry.etag == etag {
```
NobodyXu, line 302: According to [mdn doc for etag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#caching_of_unchanged_resources), we could include `None-Match: "${etag}"` in the request header, and the server will return [304](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304) Not Modified status, without a body.

tools/manifest.sh
```diff
@@ -10,11 +10,11 @@ cd "$(dirname "$0")"/..
 #    ./tools/manifest.sh [PACKAGE [VERSION_REQ]]
 
 if [[ $# -gt 0 ]]; then
-    cargo run --manifest-path tools/codegen/Cargo.toml --release -- "$@"
+    cargo +nightly run --manifest-path tools/codegen/Cargo.toml --release -- "$@"
```
NobodyXu, line 13: I know that the manifest CI has been using nightly, just curious why nightly is used cc @taiki-e 

If it is switched to stable, then I can cache it completely using [action Swatinem/rust-cache](https://github.com/Swatinem/rust-cache), saving about 30s.

NobodyXu (changes requested):

tools/codegen/src/main.rs
NobodyXu: Aha sorry I didn't read the description

> Using If-None-Match header would also work, but that would add complexity to download(..)

Hmmm that's true... Though I think we can create a new `download_if_not_cached` function to reduce the code complexity?

But if you think it's too much work, feel free to close this one, reducing 4m to 2m is already a lot.

> I found using head doesnt work, as the Server responded to ureq::head with 401 if If-None-Match is used; however responds to wget head with 304.

Perhaps the web server reacts differently to different user agent.

> as the asset fetches because they are so numerous and voluminous to be likely to prevent features being developed by new contributors who will balk at downloading so much data just to check a new feature against the full set of manifests.

I think the refetching is done to verify the data, which makes sense, though does make CI longer.

One thing we could do with that, is to checked for yank crates and remove them from the action.

tools/manifest.sh
jayvdb: It doesnt compile with stable.

tools/codegen/src/main.rs
jayvdb: the docs for github are clear that asset etags are not weak validators.

tools/codegen/src/main.rs
jayvdb: `download_if_not_cached` would be a lot of duplication.  And because github assets dont allow HEAD and etags with the ureq default user agent, we would need to make it github specific.  Might be worth it if we needed that, but we dont need it for assets.  This PR is effectively doing a HEAD and disconnecting if the etag in the header matches, before streaming the asset body.  We are doing the `If-None-Match` on the client side.

Removing yanked crates from the manifest is a great idea - but it is orthogonal to this feature, so should be a new issue.

CI only refreshes the latest version, to validate them.

tools/manifest.sh
NobodyXu: Thanks, I see the use of `LazyCell` makes it a problem.

IIRC it's going to be stablised in 1.80, and I think it makes more sense to switch to stable so that it can be cached.

NobodyXu (approved):
Thank you!

It would just need to resolve the merge conflict, which is probably faster to re-generate the manifest.

NobodyXu (dismissed):

taiki-e (commented):
Thanks for working on this!

> As a result the manifest CI job goes from 4m30s to ~2m.

In the latest run, the run time of codegen in main and this PR is only about 20 seconds different. ([main 2m48s](https://github.com/taiki-e/install-action/actions/runs/9597902868/job/26467965724), [this PR is 2m27s](https://github.com/taiki-e/install-action/actions/runs/9598801774/job/26471042273?pr=547))

That said, I think there is enough advantage on performance even CI use cases alone, as the implementation in main is more likely to be slower depend on network conditions.

> This doesnt reduce the number of API hits

The number of API hits may be increased for the second and subsequent runs in local, because the header is always fetched regardless of whether or not the cached file is present.

That said, if it allows us to detect and handle cases where the cached file is outdated, that might be good.

tools/codegen/src/main.rs
```diff
@@ -287,19 +285,40 @@ fn main() -> Result<()> {
...
+                        eprintln!("existing etag matched");
+                        download_info.insert(platform, entry.clone());
+                        continue;
+                    }
+                    eprintln!("warn: existing etag no longer valid.");
```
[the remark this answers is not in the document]
jayvdb: probably?
I would like to understand github's assets etag a bit more before doing that.
While the etag being the same guarantees that the content hasnt changed,
If the etag changes, that has no specified meaning.
And now that I have downloaded all of the files, I don't want the tool silently deleting them all because github issued new etags for the same files.
If you agree, I can create an issue to run a full download of all assets again in a month and see if any etag changed.

Also if an etag changes, I would like to keep the old file, so I can inspect the old and new files.  Getting the old file might be difficult; the saved file on my disk might be the only way I could investigate why the github project uploaded a new asset easily by comparing the old and new.

i.e. I would prefer this warning to be replaced with a hard exit, but I didn't want to do that immediately.

jayvdb:
Note this PR is broken by every manifest update (and is broken now), and it is not trivial to update it, because the new manifest cant be parsed by serde because the etag is missing.

NobodyXu (approved):
I think this PR looks good and can be merged now, to avoid more merge conflicts

taiki-e:
Sounds good, let's merge this and try it out.

PR #553, opened 2024-06-21
Add knope

jayvdb:
https://github.com/knope-dev/knope - Rust changelog manager


taiki-e:
It seems the latest version is 0.16.2, but not in the manifest.
https://github.com/knope-dev/knope/releases/tag/knope/v0.16.2

NobodyXu (approved):
Thanks, it LGTM

It doesn't have anything before 0.16.1, but I think it's not an issue

PR #554, opened 2024-06-21
Speedup codegen by caching tools/codegen compilation

NobodyXu:
- Use `once_cell::sync::Lazy` instead of `std::sync::LazyLock`, I've left a TODO comment there since it will be stablised in 1.80 on 25 July 2024
 - Switch to stable rust toolchain
 - Use [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) to cache tools/codegen

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,6 +158,8 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
+    env:
+      RUSTFLAGS: -C link-arg=-fuse-ld=lld
     steps:
       - uses: taiki-e/checkout-action@v1
       - name: Install Rust
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,7 +166,7 @@ jobs:
         run: cargo update
       - uses: Swatinem/rust-cache@v2
         with:
-          cache-all-crates: "true"
+          cache-all-crates: 'true'
       - run: tools/manifest.sh
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,7 +161,12 @@ jobs:
     steps:
       - uses: taiki-e/checkout-action@v1
       - name: Install Rust
-        run: rustup toolchain add nightly --no-self-update && rustup default nightly
+        run: rustup toolchain add stable --no-self-update --profile minimal && rustup default stable
+      - name: Generate Cargo.lock
+        run: cargo update
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-all-crates: "true"
       - run: tools/manifest.sh
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


NobodyXu:
Hm, it's not allowed to run

.github/workflows/ci.yml
```diff
@@ -161,7 +161,12 @@ jobs:
     steps:
       - uses: taiki-e/checkout-action@v1
       - name: Install Rust
-        run: rustup toolchain add nightly --no-self-update && rustup default nightly
+        run: rustup toolchain add stable --no-self-update --profile minimal && rustup default stable
+      - name: Generate Cargo.lock
+        run: cargo update
```
taiki-e, line 166: Does this make the cache invalid every time a dependency is updated?

tools/codegen/src/main.rs
```diff
@@ -673,7 +675,8 @@ impl GitHubTokens {
         }
     }
 }
-static GITHUB_TOKENS: LazyLock<GitHubTokens> = LazyLock::new(|| {
+// TODO: Use std::sync::LazyLock once 1.80 is released
+static GITHUB_TOKENS: Lazy<GitHubTokens> = Lazy::new(|| {
```
taiki-e, line 679: I'd like to use OnceLock instead adding new dep (invert of https://github.com/taiki-e/cargo-config2/commit/614d68ef65566839aa7b5fec4e9b2eb5ba824e65).

.github/workflows/ci.yml
NobodyXu: Yes, on every patch release it is invalidated, though it would still reuse the previous cache.

If the rustc version is changed, then it would start afresh.

tools/codegen/src/main.rs
NobodyXu: Sure, I will do it tomorrow (it's late here).

.github/workflows/ci.yml
```diff
@@ -161,7 +161,12 @@ jobs:
     steps:
       - uses: taiki-e/checkout-action@v1
       - name: Install Rust
-        run: rustup toolchain add nightly --no-self-update && rustup default nightly
+        run: rustup toolchain add stable --no-self-update --profile minimal && rustup default stable
```
taiki-e, line 164:
```suggestion
        run: rustup update stable --no-self-update
```

stable is already the default toolchain, and minimal is the default profile AFAIK

taiki-e:
> Hm, it's not allowed to run

Added Swatinem/rust-cache to allowed actions.

.github/workflows/ci.yml
NobodyXu: BTW it will also start afresh if `RUSTFLAGS` or any `CARGO*` env has changed.

tools/codegen/src/main.rs
NobodyXu: Updated.

.github/workflows/ci.yml
```diff
@@ -83,6 +83,8 @@ jobs:
       - uses: ./
         with:
           tool: ${{ steps.tool-list.outputs.tool }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
taiki-e, line 87: The fact that cargo-binstall's performance or robustness would be worse without this env var at a high probability is very unfortunate.

NobodyXu: Yeah, that's something I would like to improve on.

Maybe we should serailise the github requests instead, that would be faster.

taiki-e (approved):
Thanks!

PR #556, opened 2024-06-22
Set GITHUB_TOKEN for cargo-binstall by default

taiki-e:
https://github.com/taiki-e/install-action/pull/554#discussion_r1649548929

> The fact that cargo-binstall's performance or robustness would be worse without this env var at a high probability is very unfortunate.

cc @NobodyXu

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -27,3 +27,4 @@ runs:
         INPUT_TOOL: ${{ inputs.tool }}
         INPUT_CHECKSUM: ${{ inputs.checksum }}
         INPUT_FALLBACK: ${{ inputs.fallback }}
+        DEFAULT_GITHUB_TOKEN: ${{ github.token }}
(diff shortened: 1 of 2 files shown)


NobodyXu (commented):
Thank you!

I've been reluctant to do this because of the implicit github-token passing.

I think we should put up documentation on this and someway to disable it?

taiki-e:
> the implicit github-token passing

Well, that's [what the actions/checkout does](https://github.com/actions/checkout/blob/692973e3d937129bcbf40652eb9f2f61becf3332/action.yml#L24), but it might indeed not be a good idea here.

NobodyXu:
I will try to improve on cargo-binstall behavior, ideally users should not have to pass the github token, unless they are really installing a lot of crates or the crate itself doesn't provide any information in manifest and binstall has to figure out itself.

I think we can at least speed it up and avoid the need of token, for crate providing package.binstall with pkg-url

NobodyXu (approved):

NobodyXu (approved):

NobodyXu:
cc @taiki-e I have resolved the merge conflicts, can we merge this in and have another release please?

PR #741, opened 2024-11-11
chore(tools): add cargo-auditable

### LINKED ISSUE #740: add to supported tools: `cargo-auditable`
I think it would be nice, to have that as a supported tool. binstall fallback is working, but it's still taking a while.

https://github.com/rust-secure-code/cargo-auditable

simonsan:
Closes #740


tools/codegen/base/cargo-auditable.json
```diff
@@ -0,0 +1,16 @@
...
+  "tag_prefix": "v",
+  "rust_crate": "${package}",
+  "asset_name": "${package}-${rust_target}.tar.xz",
+  "bin": "${package}-${rust_target}/${package}${exe}",
+  "platform": {
+    "x86_64_linux_musl": {},
+    "x86_64_macos": {},
+    "x86_64_windows": {
+      "asset_name": "${package}-${rust_target}.zip"
+    }
```
taiki-e, line 14:
```suggestion
    },
    "aarch64_macos": {}
```

cargo-auditable also provides aarch64 macos binary.

tools/codegen/base/cargo-auditable.json
```diff
@@ -0,0 +1,16 @@
+{
+  "repository": "https://github.com/rust-secure-code/cargo-auditable",
+  "website": "https://github.com/rustsec/rustsec/tree/HEAD/cargo-audit",
```
taiki-e, line 3:
```suggestion
```

(copy-paste error?)

taiki-e (changes requested):
Thanks!

tools/codegen/base/cargo-auditable.json
simonsan: Indeed! Thanks!

simonsan:
Regenerated the manifest, should be ready to go. 👍🏽

taiki-e:
It seems to be failing in Windows.

https://github.com/taiki-e/install-action/actions/runs/11779914031/job/32809380176?pr=741#step:8:51

```
info: installing cargo-auditable@latest
info: downloading https://github.com/rust-secure-code/cargo-auditable/releases/download/v0.6.4/cargo-auditable-x86_64-pc-windows-msvc.zip
info: verifying sha256 checksum for cargo-auditable-x86_64-pc-windows-msvc.zip
caution: filename not matched:  cargo-auditable-x86_64-pc-windows-msvc/cargo-auditable.exe
```

(I think failures in test-container is an issue that needs to be addressed on my end.)

simonsan:
Should be fixed with the latest commit, it seems the folder hierarchy is different in the zip than the other one's ...

taiki-e:
> (I think failures in test-container is an issue that needs to be addressed on my end.)

I found out that this is actually not a problem on our end, but because cargo-auditable does not support --version flag and tries to forward the input to Cargo in all cases.

For example, in the following case, the output shows the version of Cargo, not the version of cargo-auditable itself.

https://github.com/taiki-e/install-action/actions/runs/11790926857/job/32842147226#step:8:72

> ```
> + cargo-auditable auditable --version
> cargo 1.82.0 (8f40fc59f 2024-08-21)
> ```

@simonsan Could you add cargo-auditable to the following branch to address this issue?

https://github.com/taiki-e/install-action/blob/f725b9adf3838dea66890ca9932712709858c409/main.sh#L792-L793

simonsan:
Added it 👍🏽

taiki-e (approved):

taiki-e:
Published in 2.45.0.

PR #757, opened 2024-11-26
Update manifest

taiki-e:
Auto-generated by CI using [create-pull-request](https://github.com/peter-evans/create-pull-request).

This will be auto-merged when CI has passed because this is an auto-generated PR in a defined format and is usually considered no additional review is required.

### DIFF
--- a/manifests/cargo-rdme.json
+++ b/manifests/cargo-rdme.json
@@ -21,6 +21,16 @@
   "1.4": {
     "version": "1.4.2"
   },
+  "1.4.8": {
+    "x86_64_linux_musl": {
+      "etag": "0x8DD0D8E946C15EC",
+      "checksum": "b3a3e141b128d3dfffe3806d9c65a001818355a515111c946775d9e2c7112ab4"
+    },
+    "x86_64_windows": {
+      "etag": "0x8DD0D8EC01F6FC1",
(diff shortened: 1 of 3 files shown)


manifests/cargo-rdme.json
```diff
@@ -21,6 +21,16 @@
   "1.4": {
     "version": "1.4.2"
   },
```
DJMcNab, line 23: The three "versions" here weren't updated.  Is this intentional, or was there some bug in the manifest generation?

DJMcNab: Hmm. That's annoying, but understandable. I was going to suggest that the install action should warn if you're in this state, but I wouldn't have seen the place where that would have been reported anyway 😅. That is, for all I know, the warning already exists.

I can't see where this is tracked upstream? Cc @orium

taiki-e: > I can't see where this is tracked upstream?

I guess it is https://github.com/orium/cargo-rdme/issues/239 / https://github.com/rust-build/rust-build.action/pull/99 (Linux, fixed), https://github.com/rust-build/rust-build.action/issues/88 (macOS)

A practical action on my part for projects with problematic release workflows is to help them switch to an action that is understandable and easy for me to set up and actually used in various parts of the ecosystem, such as in the tarpaulin case (https://github.com/taiki-e/install-action/pull/42#issuecomment-1372184541, https://github.com/xd009642/tarpaulin/pull/1174).

PR #794, opened 2024-12-23
chore(tools): add wash

joonas:
We're big fans of your [`install-action` over at the wasmCloud project](https://github.com/search?q=repo%3AwasmCloud%2FwasmCloud%20install-action&type=code).

While the `cargo-binstall` fallback does work, I'd love to be able to make our CLI (`wash`) available directly via the `install-action`, so I went ahead and added it.

In case you're wondering about the naming (`wash-cli` vs. `wash`), we're looking to move to [`wash`](https://github.com/wasmCloud/wasmCloud/issues/3870) in the near future, which aligns with the name of the tool itself.


tools/codegen/base/wash.json
```diff
@@ -0,0 +1,13 @@
+{
+  "repository": "https://github.com/wasmCloud/wasmCloud",
+  "tag_prefix": "wash-cli-v",
```
NobodyXu, line 3: Are you sure this is correct?

I don't think wasmCloud has a tag prefix?

https://github.com/wasmCloud/wasmCloud/releases/download/v1.4.2/wasmcloud-aarch64-apple-darwin

NobodyXu (changes requested):
Thanks!

tools/codegen/base/wash.json
taiki-e: I guess this pr is about wash (wash-cli), not wasmcloud.

joonas: The intent was indeed to add `wash` (the "wasmCloud shell" CLI tool), because it has plenty of uses in GitHub Actions, not wasmCloud itself 🙂

NobodyXu: Thanks for clarifying!

NobodyXu (approved):

taiki-e (approved):

taiki-e:
Published in 2.47.0.

joonas:
> Published in 2.47.0.

that's fantastic, thank you! 🎉

PR #868, opened 2025-02-23
Allow installing pre-release versions using binstall

### LINKED ISSUE #858: Installing release candidates is not supported
i think the problem itself is well understood:

> Error: install-action v2 does not support semver pre-release and build-metadata: '0.2.0-rc1'; if you need these supports again, please submit an issue at https://github.com/taiki-e/install-action

Here's our use case:

We're using this action for installing tools for testing packages in the Typst ecosystems. Since Typst is pre-1.0, there are still often breaking changes that the package ecosystem has to adapt to, and being able to easily and early test when a Typst prerelease is available is thus important, to encourage package authors to be proactive.

A PR for supporting this was made here: https://github.com/typst-community/typst-package-template/pull/31. There's an action `typst-setup` due to which Typst itself is not affected, but the testing tool `tytanic` is installed via `install-action`, and it reasonable only supports the Typst RC in a RC of its own. The workaround is to use the `cache-cargo-install-action`, but this is slower (before caching) and requires explicit support, instead of simply working with our existing setup.

So in short: we want to test packages on release candidates, the tool doing so supports RCs only on its own RCs, and if `install-action` supported RCs we wouldn't have to conditionally use different actions based on whether we're working with an RC.

I'd really appreciate it if it's not super hard to (re?) add support for this, thanks!

taiki-e:
Closes #858

cc @SillyFreak: Could you confirm that this actually solves your problem? (This PR can be used as `uses: taiki-e/install-action@pre`.)

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -639,11 +639,11 @@ for tool in "${tools[@]}"; do
   if [[ "${tool}" == *"@"* ]]; then
     version="${tool#*@}"
     tool="${tool%@*}"
-    if [[ ! "${version}" =~ ^([1-9][0-9]*(\.[0-9]+(\.[0-9]+)?)?|0\.[1-9][0-9]*(\.[0-9]+)?|^0\.0\.[0-9]+)$|^latest$ ]]; then
+    if [[ ! "${version}" =~ ^([1-9][0-9]*(\.[0-9]+(\.[0-9]+)?)?|0\.[1-9][0-9]*(\.[0-9]+)?|^0\.0\.[0-9]+)(-[0-9A-Za-z\.-]+)?$|^latest$ ]]; then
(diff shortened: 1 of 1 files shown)


SillyFreak:
Looks good! https://github.com/typst-community/typst-package-template/actions/runs/13485913904/job/37677016870?pr=31

Excerpts from the log:
```
Run taiki-e/install-action@pre
  with:
    tool: just,tytanic@0.2.0-rc1

 INFO resolve: Resolving package: 'tytanic@=0.2.0-rc1'
 WARN The package tytanic v0.2.0-rc1 (x86_64-unknown-linux-musl) has been downloaded from github.com
 INFO This will install the following binaries:
 INFO   - tt => /home/runner/.cargo/bin/tt
```

```
Run tt --version

  tt --version
  just test
  shell: /usr/bin/bash -e {0}
tytanic 0.2.0-rc1
tt run 
  Starting 2 tests (run ID: 465f3ea3-f507-438e-b411-f04626872c19)
      pass [     48ms] test1
   compile [      7ms] unit1
──────────
   Summary [     55ms] 2/2 tests run: all 2 passed
```

so yes, the action installs the prerelease, the binary is actually the expected one, and it runs as expected.

taiki-e:
Thanks for confirming!

taiki-e:
Published in 2.49.0.

Issue #971, opened 2025-05-19
Fetching head of licenses failing CI


jayvdb:
https://github.com/taiki-e/install-action/actions/runs/15107664605/job/42459736285?pr=970

```
     Running `target/release/generate-manifest cargo-dinghy latest`
download cache: /home/runner/work/install-action/install-action/tools/codegen/tmp/cache/cargo-dinghy
downloading metadata from https://api.github.com/repos/sonos/dinghy
downloading releases from https://api.github.com/repos/sonos/dinghy/releases
downloading crate info from https://crates.io/api/v1/crates/cargo-dinghy
Trying to using license 'MIT/Apache-2.0' from crates.io ...
fetching head of https://raw.githubusercontent.com/sonos/dinghy/main/LICENSE-MIT ..
fetching head of https://raw.githubusercontent.com/sonos/dinghy/main/LICENSE-APACHE ..

thread 'main' panicked at tools/codegen/src/main.rs:859:17:
Unable to find license files in the repo for all licenses [(MIT, None), (Apache-2.0, None)]; found ["[Apache-2.0](https://github.com/sonos/dinghy/blob/main/LICENSE-APACHE)"]
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: generate_manifest::get_license_markdown
   3: generate_manifest::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
manifest.sh:21: `cargo run --manifest-path tools/codegen/Cargo.toml --release -- "${package}" latest` exit with 101
```

I do `git commit --amend -a` and push and then

https://github.com/taiki-e/install-action/actions/runs/15110602941/job/42469037819?pr=970
```
     Running `target/release/generate-manifest cargo-deny latest`
download cache: /home/runner/work/install-action/install-action/tools/codegen/tmp/cache/cargo-deny
downloading metadata from https://api.github.com/repos/EmbarkStudios/cargo-deny
downloading releases from https://api.github.com/repos/EmbarkStudios/cargo-deny/releases
downloading crate info from https://crates.io/api/v1/crates/cargo-deny
Trying to using license 'MIT OR Apache-2.0' from crates.io ...
fetching head of https://raw.githubusercontent.com/EmbarkStudios/cargo-deny/main/LICENSE-MIT ..
fetching head of https://raw.githubusercontent.com/EmbarkStudios/cargo-deny/main/LICENSE-APACHE ..

[...]

PR #973, opened 2025-05-20
Check existing license_markdown before license detection fallback

jayvdb:
This change adds checking of the `license_markdown`, whether it was in the base, or previous detected & generated.

It only falls back to detecting the licenses if that existing `license_markdown` contained URLs that are no longer able to be fetched.  This reduces the amount of work needed per tool.  Also it uses the GitHub API to fetch the license files, instead of the Github Raw file URLs, as the latter has rate limits tied to the IP address which in CI is less reliable, while the former is tied to the token used and is more predictable.

As a result, it does mean the tool SPDX license expression could change, and not be noticed if the license files did not change in the repository.  Examples of how that could occur:
- "MIT AND Apache-2.0" becomes "MIT OR Apache-2.0"
- MIT pointing to file `LICENSE` becomes Apache-2.0 pointing to file `LICENSE`

Closes https://github.com/taiki-e/install-action/issues/971

### DIFF
--- a/tools/codegen/src/main.rs
+++ b/tools/codegen/src/main.rs
@@ -127,7 +127,9 @@ fn main() -> Result<()> {
             latest_only = true;
         }
     }
+
     if manifest_path.is_file() {
+        println!("loading pre-existing manifest {}", manifest_path.display());
         match serde_json::from_slice(&fs::read(manifest_path)?) {
             Ok(m) => {
                 manifests = m;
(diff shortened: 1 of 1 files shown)


tools/codegen/src/main.rs
```diff
@@ -161,28 +163,44 @@ fn main() -> Result<()> {
...
+                }
+            }
+        } else if let Some(license) = repo_info.license {
+            if let Some(license) = license.spdx_id {
+                eprintln!("Trying to verify license '{license}' obtained from github.com ...");
+                if let Some(license_markdown) =
+                    get_license_markdown(&license, repo, &repo_info.default_branch)
+                {
+                    manifests.license_markdown = license_markdown;
+                }
```
NobodyXu, line 203: Isn't the two branches doing the same thing, despite having different logging?

jayvdb: yes, and this logic hasnt changed in this PR; the lines have been indented, and the stderr output message has been slightly improved to be more readable.

jayvdb: Overhauled this to do the destructuring more neatly and thus DRY up that inner bit of code.

NobodyXu (approved):
Thanks, LGTM

taiki-e:
Thanks for the fix! This was a problem that CI had been failing frequently in the recent past.

PR #980, opened 2025-05-22
Update `cargo-sort@latest` to 2.0.0

jplatte:
I ran `./tools/manifest.sh cargo-sort`. Looks like `macos-latest` GitHub runners are now aarch64, and that script isn't detecting the aarch64 binary that was uploaded to the [release](https://github.com/DevinR528/cargo-sort/releases/tag/v2.0.0).

PR #997, opened 2025-06-11
Update cargo-sort v2.x replacing x86_64_macos with aarch64_macos

jayvdb:
Replaces https://github.com/taiki-e/install-action/pull/980


taiki-e:
Thanks! CI failure on test-manifest-schema has been fixed, so I think if the error regarding the GLIBC version is addressed, it will be possible to merge.

PR #1000, opened 2025-06-15
Add cyclonedx

jayvdb:
https://github.com/CycloneDX/cyclonedx-cli


jayvdb:
Interesting
```
+ file /home/runner/.install-action/bin/cyclonedx
/home/runner/.install-action/bin/cyclonedx: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, BuildID[sha1]=ffd8fa9153d9deefccbf7341ab040da3620fb0ea, stripped
+ ldd /home/runner/.install-action/bin/cyclonedx
	linux-vdso.so.1 (0x00007fff1b7f4000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f1cbdfd4000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f1cbdfb4000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f1cbd000000)
	libc.musl-x86_64.so.1 => not found
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1cbcc00000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f1cbdecb000)
	/lib/ld-musl-x86_64.so.1 => /lib64/ld-linux-x86-64.so.2 (0x00007f1cbdffe000)
```

jayvdb:
Now failing with
```
Process terminated. Couldn't find a valid ICU package installed on the system. Please install libicu using your package manager and try again. Alternatively you can set the configuration flag System.Globalization.Invariant to true if you want to run with no globalization support. Please see https://aka.ms/dotnet-missing-libicu for more information.
```

https://github.com/CycloneDX/cyclonedx-cli/blob/f934c99826339cb8dbb83b439eb2c465fb253fb3/README.md?plain=1#L346 says the following are needed.

> For Ubuntu these are libc6 libgcc1 libgssapi-krb5-2 libicu66 libssl1.1 libstdc++6 zlib1g.

tools/manifest-schema/src/lib.rs
```diff
@@ -282,6 +282,17 @@ impl BaseManifest {
...
+            panic!("At least one platform must be specified");
+        }
+        if !self.prefer_linux_gnu
+            && self.platform.contains_key(&HostPlatform::x86_64_linux_gnu)
+            && self.platform.contains_key(&HostPlatform::x86_64_linux_musl)
+        {
+            panic!(
+                "When x86_64-linux-musl platform is specified, x86_64-linux-gnu will never be used and should not be specified"
+            );
+        }
```
taiki-e, line 295: This PR currently only checks x86_64_linux but why doesn't the same check apply to aarch64_linux?

tools/manifest-schema/src/lib.rs
```diff
@@ -282,6 +282,19 @@ impl BaseManifest {
...
+            panic!("At least one platform must be specified");
+        }
+        if !self.prefer_linux_gnu
+            && (self.platform.contains_key(&HostPlatform::x86_64_linux_gnu)
+                && self.platform.contains_key(&HostPlatform::x86_64_linux_musl))
+            && (self.platform.contains_key(&HostPlatform::aarch64_linux_gnu)
+                && self.platform.contains_key(&HostPlatform::aarch64_linux_musl))
+        {
+            panic!(
+                "When x86_64-linux-musl platform is specified, x86_64-linux-gnu will never be used and should not be specified"
```
taiki-e, line 295:
```suggestion
                "When *-linux-musl platform is specified, *-linux-gnu for the same architecture will never be used and should not be specified"
```

taiki-e:
Published in 2.54.0.

PR #1016, opened 2025-06-29
Add vacuum

jayvdb:
https://github.com/daveshanley/vacuum

Rust currently doesnt have a good openapi v3.1 parser , let alone linter.

( https://github.com/blst-security/cherrybomb is reasonable, not as good as vacuum , but is openapi v3.0 only)


taiki-e:
Published in 2.55.0.

PR #1019, opened 2025-06-30
Update zola install location in TOOLS.md

jayvdb:
This is a side-effect of https://github.com/taiki-e/install-action/pull/932


taiki-e (approved):
Thanks.

PR #1060, opened 2025-07-26
feat: Add `mdbook-alerts` as a new tool

CommanderStorm:
First of all, thanks for this wonderfull thing you did.

This PR adds `mdbook-alerts` as a new tool.
The [Github releases can be found here](https://github.com/lambdalisue/rs-mdbook-alerts/releases).

In terms of verifcation, I have
- checked that the urls match (= can download) for both linux, windows, macos,
- that docs.rs points to said repo 
- that the URLs in TOOLS.md are correct
- platforms match what is offerered
- licence matches


taiki-e (approved):
Thanks!

taiki-e:
Published in 2.57.0.

PR #1073, opened 2025-08-05
Add iai-callgrind-runner

niklasf:
https://github.com/iai-callgrind/iai-callgrind is a benchmark harness that comes as a dev dependency plus a companion tool. The latter is useful to quickly install in CI.

Tested with:

```yaml
uses: niklasf/install-action@iai-callgrind-runner
with:
    tool: iai-callgrind-runner@0.16.1,valgrind
```


taiki-e (approved):
Thanks!

taiki-e:
Published in 2.58.0.

Issue #1100, opened 2025-08-16
`taplo` always falls back on `cargo-binstall` for installation


AudaciousAxiom:
First, thank you for your work on this Action!

It seems that, even though support for [taplo/taplo-cli](https://github.com/tamasfe/taplo) has been added in #944, `install-action` still always falls back on `cargo-binstall` to install it.
See for instance [this CI run of `rustup`](https://github.com/rust-lang/rustup/actions/runs/17009891144/job/48224498831) where `install-action` prints the following:

```sh
info: install-action does not support taplo-cli; fallback to cargo-binstall
```

[Here is how `rustup`'s CI attempts to use `install-action` to install `taplo-cli` directly.](https://github.com/rust-lang/rustup/blob/987c185a6873efbdfc0b85345cddee2189d7cdf3/.github/workflows/ci.yaml#L1464-L1467)

I have been able to reproduce this both on GitHub and Codeberg, and this does not seem to be a networking issue, as it is not transient.

Additionally, this does not seem to originate from a confusion between `taplo` and `taplo-cli`, as I have tried both to no avail.

I am not sure how to investigate this issue further.

PR #1128, opened 2025-09-05
Add taplo-cli alias

arctic-alpaca:
If I understood https://github.com/taiki-e/install-action/issues/1100#issuecomment-3193786729 correctly, this should be all that's needed to add an alias for `taplo-cli`.


taiki-e (approved):
Thanks!

PR #1135, opened 2025-09-09
Add ubi

jayvdb:
https://github.com/houseabsolute/ubi seems to be the best Rust equivalent of https://github.com/zyedidia/eget


taiki-e (approved):
Thanks!

Issue #515, opened 2024-06-08
Feature request: Support installation of ziglang


NobodyXu:
My understanding of this action is that it does not do extra installs that might be required to make a tool work.
e.g. installing `cargo-valgrind` does not also install `valgrind` ; installing `xbuild` does not also install `adb` and other bits of llvm not present on the github runner.

It is up to the user of this action to install dependencies that are missing, and that way they can choose the version they want of those dependencies.

I would support adding an extra tool called "ziglang", which would be a special case also like "valgrind".  We could support installing different versions from https://pypi.org/project/ziglang/#history - that looks like a reasonable approach, but Python wheels are a messy format for this given that `zig` is a single binary, and pip is slow.  Importing the binary URLs from https://ziglang.org/download/index.json would be more in tune with the way this action works, allowing runtime verification of the checksums.

_Originally posted by @jayvdb in https://github.com/taiki-e/install-action/issues/508#issuecomment-2150934054_

jayvdb:
mise can install zig

https://mise.jdx.dev/lang/zig.html#zig

jayvdb:
In my CI I was able to get zig installed using

```sh
mise settings add idiomatic_version_file_enable_tools "[]"
mise use --global zig
echo "$HOME/.local/share/mise/shims" >> "$GITHUB_PATH"
echo "Added $HOME/.local/share/mise/shims to PATH"
```

taiki-e:
FYI, https://github.com/taiki-e/install-action/pull/1779 added support for installation of Rust.

PR #1137, opened 2025-09-09
Add mise

jayvdb:
https://github.com/jdx/mise , which can be used to solve https://github.com/taiki-e/install-action/issues/515


tools/codegen/base/mise.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "x86_64_windows": {
+      "asset_name": "${package}-v${version}-${rust_target_os}-x64.zip"
+    },
+    "aarch64_linux_musl": {
+      "asset_name": "${package}-v${version}-${rust_target_os}-arm64-musl.tar.gz"
+    },
+    "aarch64_macos": {
+      "asset_name": "${package}-v${version}-${rust_target_os}-arm64.tar.gz"
+    }
```
taiki-e, line 22: I think we can add the following platform:
- aarch64_windows: mise-v2025.9.7-windows-arm64.zip

taiki-e (approved):
Thanks!

PR #1364, opened 2025-12-17
feat: add `martin`

CommanderStorm:
this is a re-attempt of 
- https://github.com/taiki-e/install-action/pull/1075


taiki-e:
Published in 2.64.0.

PR #1647, opened 2026-04-01
Call main.sh from powershell on Windows to work around windows-11-arm runner bug

### LINKED ISSUE #1472: No package installed with no output on windows-arm runner
Never encountered this before, not sure a GHA issue or something else, but I'd like to post it here and see if others encountered similar issues

https://github.com/cargo-bins/cargo-binstall/actions/runs/21423048835/job/61686304252#step:4:165

### LINKED ISSUE #1562: On windows-11-arm, sometimes does nothing, but "succeeds"
I doubt this is an issue with this action, maybe some hiccup in GitHub actions' infrastructure, not sure, but asking here just in case you have any ideas 🙏.

Sometimes this action shows as green, but logs nothing, and when we try to use the tool we asked install-action to install, it's not there. It's specific to windows-11-arm and happens just sporadically.

You can see an example here: https://github.com/spinel-coop/rv/actions/runs/22726945016/job/65904800032?pr=584.

Thanks so much, feel free to close if definitely not this actions' fault.

### LINKED ISSUE #1645: Silently doesn't install sometimes
This happened a few times for me, the latest occurrence is here: https://github.com/nazar-pc/abundance/actions/runs/23842313355/job/69500910852?pr=605

Essentially, the action completed successfully, but `cargo-nextest` wasn't actually installed in practice.

It should probably error if installation was not possible for whatever reason.

taiki-e:
Workaround for windows-11-arm runner bug: https://github.com/actions/partner-runner-images/issues/169.
Fixes #1645
Fixes #1562
Fixes #1472

The issue appears to be bash sometimes failing to start either due to how GHA launches bash or regardless of who to launch bash.

In the former case, just calling bash from powershell should be sufficient (as currently implemented in this PR); in the latter case, I think it would be sufficient to add a way to detect whether bash failed to start (such as creating and checking a temporary file) and then retry a few times.

I haven't encountered this issue in my repositories yet, so it's difficult for me to test whether the former approach is sufficient. (For now, I plan to continually rerun the tests using the windows-11-arm runner in this PR.)

I would appreciate it if anyone who has encountered this issue could help with testing.
cc @wmmc88 @nazar-pc @NobodyXu @deivid-rodriguez

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -45,8 +45,25 @@ runs:
     - run: bash --noprofile --norc "${GITHUB_ACTION_PATH:?}/main.sh"
       shell: bash
       env:
+        # NB: Sync with non-Windows case.
         INPUT_TOOL: ${{ inputs.tool }}
         INPUT_CHECKSUM: ${{ inputs.checksum }}
         INPUT_FALLBACK: ${{ inputs.fallback }}
         DEFAULT_GITHUB_TOKEN: ${{ github.token }}
         ACTION_USER_AGENT: ${{ github.action_repository }} (${{ github.action_ref }})
+      if: runner.os != 'Windows'
+    # Workaround for https://github.com/actions/partner-runner-images/issues/169
+    # TODO: Is it necessary to retry for main.sh call? Or is this sufficient? https://github.com/taiki-e/install-action/pull/1647
+    - run: |
+        Set-StrictMode -Version Latest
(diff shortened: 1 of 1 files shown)


taiki-e:
Thanks for testing this and reporting the issue! I'll implement retry (and logging).

PR #1657, opened 2026-04-02
Retry on bash startup failure on Windows

taiki-e:
Closes #1656
See that issue for details.

I've tested failure cases by modifying main.sh:
- Case that always fails to startup: https://github.com/taiki-e/install-action/actions/runs/23897413167
- Case that sometimes fails to startup/install: https://github.com/taiki-e/install-action/actions/runs/23897345379
- Case that sometimes fails to startup: https://github.com/taiki-e/install-action/actions/runs/23897548479

### DIFF
--- a/.github/.cspell/project-dictionary.txt
+++ b/.github/.cspell/project-dictionary.txt
@@ -17,6 +17,7 @@ grcov
 gungraun
 insta
 knope
+LASTEXITCODE
 libicu
 linkcheck
 mdbook
@@ -38,6 +39,7 @@ sigstore
 syft
 tombi
 udeps
+USERPROFILE
 wasmtime
 watchexec
 worktree
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Note: In this file, do not use the hard wrap in the middle of a sentence for com
 
 ## [Unreleased]
 
+- Implement workaround for [windows-11-arm runner bug](https://github.com/actions/partner-runner-images/issues/169) which sometimes causes installation failure. ([#1657](https://github.com/taiki-e/install-action/pull/1657))
+
+  This addresses an issue that was attempted to be worked around in 2.71.0 but was insufficient.
+
 - Update `mise@latest` to 2026.4.1.
 
 - Update `uv@latest` to 0.11.3.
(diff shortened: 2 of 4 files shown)


Darksonn:
Is `taiki-e/install-action@v2` enough to get this fix? I'm still seeing this.

nazar-pc:
Haven't seen this issue since updating to 2.71.2

taiki-e:
@Darksonn 
Thanks for the report! I checked the log:

- This PR's fix is included in the version [you're using](https://github.com/tokio-rs/tokio/actions/runs/24440222967/job/71403445769#step:1:27) (eea29cff9a2b68892c0845ae3e4f45fc47ee9354).
- The [issue you encountered occurs after bash has started](https://github.com/tokio-rs/tokio/actions/runs/24440222967/job/71403445769#step:4:60). (This PR fixed the issue with bash startup.)
- The issue you encountered seems to indicate that the windows-11-arm runner bug also exists when starting x86_64 binaries other than bash. In both cases, the root cause of the issue seems to be the same. 
  - In the [log](https://github.com/tokio-rs/tokio/actions/runs/24440222967/job/71403445769#step:4:60), [cargo location](https://github.com/taiki-e/install-action/blob/eea29cff9a2b68892c0845ae3e4f45fc47ee9354/main.sh#L574) is displayed, but ["installing "](https://github.com/taiki-e/install-action/blob/eea29cff9a2b68892c0845ae3e4f45fc47ee9354/main.sh#L828) is not, so I think the startup of the command is failing due to one of dirname ([1](https://github.com/taiki-e/install-action/blob/eea29cff9a2b68892c0845ae3e4f45fc47ee9354/main.sh#L575), [2](https://github.com/taiki-e/install-action/blob/eea29cff9a2b68892c0845ae3e4f45fc47ee9354/main.sh#L597)) or jq ([1](https://github.com/taiki-e/install-action/blob/eea29cff9a2b68892c0845ae3e4f45fc47ee9354/main.sh#L688-L689), [2](https://github.com/taiki-e/install-action/blob/eea29cff9a2b68892c0845ae3e4f45fc47ee9354/main.sh#L239), and more).
    - dirname is likely installed in the same way as bash, so it’s the most suspicious.

I removed the use of dirname in 2.75.14 (https://github.com/taiki-e/install-action/commit/6fc852a7b207369c7375466f6742d9d77030dead), so hopefully that will resolve or mitigate the issue you encountered.

PR #1731, opened 2026-04-19
Support `cargo-apple-runner`

madsmtm:
https://github.com/madsmtm/cargo-apple-runner, it's a Cargo runner I'm developing for making it easy to run `cargo test --target aarch64-apple-ios-sim`, which is especially useful in CI.


madsmtm:
Hmm, while I _could_ publish precompiled GH releases for Linux too, this tool currently only really makes sense on macOS.

Your CI seems to have trouble with that though, do you have a suggestion for how to fix that?

taiki-e:
Thanks for the PR! 

The CI issue is caused by our test script assuming that everything is available on x86_64_linux, so I think it should be fixed by removing `[[ "${host_arch}" != "x86_64" ]] && ` in the following line.

https://github.com/taiki-e/install-action/blob/6c30010265446c48ca58ba69185adf76b78476b4/tools/ci/tool-list.sh#L151

scutuatua-crypto (approved):

scutuatua-crypto (approved):

madsmtm:
Hmm, that seems to error with:
```
tools/ci/tool-list.sh: line 151: jq: command not found
```

I added an explicit check for `cargo-apple-runner` for now, but later on you might want to fix it by installing `jq` earlier (instead of in `main.sh`)?

taiki-e (approved):
Thanks!

taiki-e:
Published in v2.76.0.

PR #1812, opened 2026-05-11
Add `cargo-mutants`

jakewimmer:
Adds cargo-mutants to the native manifest system. Replaces the cargo-binstall fallback with direct binary downloads.

## Changes

- `tools/codegen/base/cargo-mutants.json`: new base manifest
- `manifests/cargo-mutants.json`: generated, covers 24 versions with download URLs and SHA256 checksums
- `TOOLS.md`: Adds `cargo-mutants` between cargo-minimal-versions and cargo-nextest
- `tools/ci/tool-list.sh`: Adds `cargo-mutants` to `glibc_pre_2_39_incompat`

## Asset naming

cargo-mutants switched archive formats at v25. Pre-v25 releases are `.tar.xz`, v25+ are `.tar.gz`. The base manifest lists both so the codegen picks the right one per version:

```json
"asset_name": ["${package}-${rust_target}.tar.gz", "${package}-${rust_target}.tar.xz"]
```

Same pattern as cargo-hack. Windows uses `.zip` as a platform override.

## Platforms

- x86_64 Linux (gnu)
- x86_64 macOS
- x86_64 Windows

Binaries require GLIBC_2.39. Cargo-mutants is skipped on older Linux runners in CI.

Tag prefix `v`, license MIT. No changes to `main.sh` as cargo-mutants supports `--version`.

Closes #1668


taiki-e (approved):
Thanks!

taiki-e:
Published in 2.78.0.

PR #1814, opened 2026-05-12
Change filename format for cargo-zigbuild.

simonhollingshead:
I noticed that the January 29th release of cargo-zigbuild changed the archive format.  This has caused the fetching pipeline to stall and no longer ingest any new version.

This change swaps the asset and binary names as required to be capable of fetching all versions between 0.21.5 and 0.22.3 (latest).

This also adds the macos ARM builds.

### DIFF
--- a/tools/codegen/base/cargo-zigbuild.json
+++ b/tools/codegen/base/cargo-zigbuild.json
@@ -3,19 +3,21 @@
   "license_markdown": "[MIT](https://github.com/rust-cross/cargo-zigbuild/blob/main/LICENSE)",
   "tag_prefix": "v",
   "rust_crate": "${package}",
-  "asset_name": "${package}-v${version}.${rust_target}.tar.gz",
-  "version_range": ">= 0.18.4",
+  "asset_name": "${package}-${rust_target}.tar.xz",
+  "version_range": ">= 0.21.5",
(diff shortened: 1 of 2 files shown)


tools/codegen/base/cargo-zigbuild.json
```diff
@@ -3,19 +3,21 @@
...
-    "x86_64_macos": {
-      "asset_name": "${package}-v${version}.apple-darwin.tar.gz"
-    },
+    "x86_64_macos": {},
     "x86_64_windows": {
-      "asset_name": "${package}-v${version}.windows-x64.zip"
+      "asset_name": "${package}-${rust_target}.zip",
+      "bin": "${package}${exe}"
     },
-    "aarch64_linux_musl": {},
```
taiki-e, line 16: Hmm. Why is the aarch64 musl binary no longer supported?

taiki-e (commented):
Thanks!

tools/codegen/base/cargo-zigbuild.json
simonhollingshead: As of 0.21.5 they swapped from building aarch64-musl to aarch64-gnu.

taiki-e (approved):

taiki-e:
Published in 2.78.3.

PR #1841, opened 2026-05-16
Always fall back to cargo install except on MIPS

Gelbpunkt:
You can see it in action in this run: https://github.com/Gelbpunkt/compiler-builtins/actions/runs/25971341594/job/76344755924 (a previous execution of the job successfully installed `cargo-nextest`)

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -598,13 +598,13 @@ case "${RUNNER_ARCH}" in
     info "unrecognized runner.arch '${RUNNER_ARCH}'; fallback to uname -m"
     case "$(uname -m)" in
       aarch64 | arm64) host_arch=aarch64 ;;
-      xscale | arm | armv*l) bail "32-bit Arm runner is currently not supported; if you need support for this platform, please submit an issue at <https://github.com/taiki-e/install-action>" ;;
       ppc64le) host_arch=powerpc64le ;;
       riscv64) host_arch=riscv64 ;;
       s390x) host_arch=s390x ;;
-      # Very few tools provide prebuilt binaries for these.
-      # TODO: fallback to `cargo install`? (binstall fallback is not good idea here as cargo-binstall doesn't provide prebuilt binaries for these.)
-      loongarch64 | mips | mips64 | ppc | ppc64 | sun4v) bail "$(uname -m) runner is not supported yet; please submit an issue at <https://github.com/taiki-e/install-action>" ;;
+      # On these platforms, we just use the result of `uname -m` as host_arch, and always fallback to `cargo install`.
+      xscale | arm | armv*l | loongarch64 | ppc | ppc64 | sun4v) host_arch="$(uname -m)" ;;
+      # Ignore MIPS for now, as we also need to detect endianness.
+      mips | mips64) bail "MIPS runner is not supported yet; please submit an issue at <https://github.com/taiki-e/install-action>" ;;
       # GitHub Actions Runner supports x86_64/AArch64/Arm Linux and x86_64/AArch64 Windows/macOS.
       # https://github.com/actions/runner/blob/v2.332.0/.github/workflows/build.yml#L24
       # https://docs.github.com/en/actions/reference/runners/self-hosted-runners#supported-processor-architectures


main.sh
```diff
@@ -591,6 +591,7 @@ case "${RUNNER_ARCH}" in
   ARM) bail "32-bit Arm runner is currently not supported; if you need support for this platform, please submit an issue at <https://github.com/taiki-e/install-action>" ;;
   X86) bail "32-bit x86 runner is currently not supported; if you need support for this platform, please submit an issue at <https://github.com/taiki-e/install-action>" ;;
   ARM64) host_arch=aarch64 ;;
+  PPC64) host_arch=powerpc64 ;;
```
taiki-e: this is not correct

https://github.com/Gelbpunkt/compiler-builtins/actions/runs/25971341594/job/76344755924#step:22:28

> info: unrecognized runner.arch 'ppc64'; fallback to uname -m


```suggestion
```

main.sh
```diff
@@ -599,12 +600,13 @@ case "${RUNNER_ARCH}" in
...
       aarch64 | arm64) host_arch=aarch64 ;;
       xscale | arm | armv*l) bail "32-bit Arm runner is currently not supported; if you need support for this platform, please submit an issue at <https://github.com/taiki-e/install-action>" ;;
+      ppc64) host_arch=powerpc64 ;;
       ppc64le) host_arch=powerpc64le ;;
       riscv64) host_arch=riscv64 ;;
       s390x) host_arch=s390x ;;
       # Very few tools provide prebuilt binaries for these.
       # TODO: fallback to `cargo install`? (binstall fallback is not good idea here as cargo-binstall doesn't provide prebuilt binaries for these.)
-      loongarch64 | mips | mips64 | ppc | ppc64 | sun4v) bail "$(uname -m) runner is not supported yet; please submit an issue at <https://github.com/taiki-e/install-action>" ;;
+      loongarch64 | mips | mips64 | ppc | sun4v) bail "$(uname -m) runner is not supported yet; please submit an issue at <https://github.com/taiki-e/install-action>" ;;
```
taiki-e: As the TODO comment says, I think it’s fine to always fall back to `cargo-install` for these except for MIPS, where the endianness cannot be determined from `uname -m`.

```suggestion
      # On these platforms, we just use the result of `uname -m` as host_arch, and always fallback to `cargo install`.
      xscale | arm | armv*l | loongarch64 | ppc | ppc64 | sun4v) host_arch="$(uname -m)" ;;
      # Ignore MIPS for now, as we also need to detect endianness.
      mips | mips64) bail "MIPS runner is not supported yet; please submit an issue at <https://github.com/taiki-e/install-action>" ;;
```

main.sh
```diff
@@ -599,12 +600,13 @@ case "${RUNNER_ARCH}" in
     case "$(uname -m)" in
       aarch64 | arm64) host_arch=aarch64 ;;
       xscale | arm | armv*l) bail "32-bit Arm runner is currently not supported; if you need support for this platform, please submit an issue at <https://github.com/taiki-e/install-action>" ;;
+      ppc64) host_arch=powerpc64 ;;
```
taiki-e:
```suggestion
```

main.sh
taiki-e: If you intend to actually support prebuilt binaries on this target, the current implementation is fine as it is. (However, due to differences in the ABI between musl and glibc, it is unclear whether we will be able to support it in the same way as other architectures.)

main.sh
Gelbpunkt: I believe this may be due to the way act (the runner I'm using) works. IIRC I haven't ever seen this be in caps on any platform

Gelbpunkt: As far as I can tell it has a hardcoded mapping at https://github.com/nektos/act/blob/123167dcaf6ec2c6a7e391ba21afadb55e4fadb5/pkg/container/docker_run.go#L267 and this is the fallback return value because it's not x86 or ARM. No idea where the PPC64LE entry is from since GitHub's official runner doesn't support ppc64le, but with act, that'd also always be lowercase at the moment

taiki-e: RISCV64 is from actual execution in https://github.com/riseproject-dev/riscv-runner-images. PPC64LE and S390X are guesses based on existing patterns and .NET's [ProcessorArchitecture Class](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.processorarchitecture?view=msbuild-18-netcore) (github action runner is written in .NET. Since compiler-builtins uses the corresponding IBM runners, if it's incorrect, I plan to fix it when the install-action is updated there.)

I would like to avoid writing based on guessing about things that cannot be confirmed.

Gelbpunkt:
Applied all your suggestions, thanks for the feedback :)

Gelbpunkt:
Hold up, I think this is incorrect though, we still need to map ppc64 to powerpc64.

Edit: I guess that would only be necessary for binary installs

taiki-e (approved):
Thanks!

taiki-e:
If there's no need to support installing pre-built binaries, this should be fine as is. If support for installing pre-built binaries is needed, changes will be required not only to `main.sh` but also to the manifest generation process (like https://github.com/taiki-e/install-action/pull/1133). Although I suspect very few tools actually support big-endian ppc64.

Gelbpunkt:
> If there's no need to support installing pre-built binaries, this should be fine as is

Yeah, but that should also apply to MIPS, or am I missing something? If we always fall back to `cargo install`, the installed toolchain determines the endianness, not install-action

taiki-e:
Published in 2.79.0.

> > If there's no need to support installing pre-built binaries, this should be fine as is
> 
> Yeah, but that should also apply to MIPS, or am I missing something? If we always fall back to `cargo install`, the installed toolchain determines the endianness, not install-action

You are right. I updated the code to allow MIPS (https://github.com/taiki-e/install-action/commit/41960a65de0ef502fbb802b784b4b4a859df727f).

PR #1844, opened 2026-05-17
Deprecate mdbook-alerts and iai-callgrind-runner

taiki-e:
- Deprecate `mdbook-alerts` because the feature now included in `mdbook` and the [repository](https://github.com/lambdalisue/rs-mdbook-alerts) has been archived.

- Deprecate `iai-callgrind-runner` because it has been [renamed](https://github.com/gungraun/gungraun/releases/tag/v0.17.0) to `gungraun-runner`. `gungraun-runner` is also supported by this action.

For now, this is just a documentation update. In the future, we may display a warning message when the action is executed.


taiki-e:
FYI @CommanderStorm who added mdbook-alerts in #1060.
FYI @niklasf who added iai-callgrind-runner in #1073.

CommanderStorm (approved):
`mdbook-alerts` can be fully removed. It is not nessary any more.

PR #1903, opened 2026-06-11
Support cargo-leptos

404Simon:


tools/codegen/base/cargo-leptos.json
```diff
@@ -0,0 +1,17 @@
+{
+  "repository": "https://github.com/leptos-rs/cargo-leptos",
+  "license_markdown": "[MIT](https://github.com/leptos-rs/cargo-leptos/blob/main/LICENSE)",
+  "tag_prefix": "v",
+  "rust_crate": "${package}",
+  "asset_name": "${package}-${rust_target}.tar.gz",
+  "prefer_linux_gnu": true,
```
taiki-e: This (prefer_linux_gnu) does not work on its own (it requires adjustments in `main.sh`). Additionally, this should not be used unless there are compatibility issues (such as musl binaries being dynamically linked and failing to run in a glibc environment) or significant performance benefits.

taiki-e (changes requested):
Thanks for the PR! 

Looking at the CI failure, the binary is likely contained in a directory within the archive, so I guess you have to use the `bin` field like:
https://github.com/taiki-e/install-action/blob/15449e3094499af05d8d964a1c884208e4b8b595/tools/codegen/base/cargo-audit.json#L8

tools/codegen/base/cargo-leptos.json
404Simon: Sorry for getting back to this just now, I kinda forgot about this PR.

taiki-e (approved):

taiki-e:
Published in 2.82.0.

PR #1924, opened 2026-07-02
feat: add cargo-about

ruffsl:
[cargo-about](https://github.com/EmbarkStudios/cargo-about) (EmbarkStudios)
> 📜 Cargo plugin to generate list of all licenses for a crate 🦀

publishes prebuilt release binaries for Linux (musl), macOS, and Windows, so install-action can fetch them from GitHub Releases instead of falling back to cargo-binstall. (The source fallback builds cargo-about without its cli feature, required-features = ["cli"], no default features, so leaving no cargo about binary.)


manifests/cargo-about.json
```diff
@@ -0,0 +1,895 @@
...
+    "aarch64_windows": {
+      "url": "https://github.com/EmbarkStudios/cargo-about/releases/download/${version}/cargo-about-${version}-aarch64-pc-windows-msvc.tar.gz",
+      "bin": "cargo-about-${version}-aarch64-pc-windows-msvc/cargo-about.exe"
+    }
+  },
+  "latest": {
+    "version": "0.9.0"
+  },
+  "0.9": {
+    "version": "0.9.0"
```
taiki-e:
```suggestion
    "version": "0.9.1"
  },
  "0.9": {
    "version": "0.9.1"
```

taiki-e (approved):
Thanks!

taiki-e:
Published in 2.83.0.

PR #1941, opened 2026-07-18
Update `convco` to 0.7.0

graelo:
Hi,

Convco 0.7.0 was released this week, with a new naming scheme for the release binaries, and a new file layout for each (binary in versioned dir).

Here's the updated manifest.

Cheers


taiki-e (approved):
Thanks!

taiki-e:
Published in 2.84.0.

PR #1966, opened 2026-08-13
Update `bpf-linker@latest` to 0.11.0

tamird:


tamird:
Ah this is also included in #1965.

taiki-e:
Published 2.85.13 which includes bpf-linker update.

PR #1967, opened 2026-08-13
feat(manifest): add `oxfmt`

rami3l:
This PR adds `oxfmt` to the set of supported tools.

## Background

Thank you very much for making this action! In the current `rustup` CI, we are already using it to install `taplo`. However, rustup also has a website, which requires consistent frontend formatting in addition to TOML style checks.

Originally requested in https://github.com/rust-lang/rustup/pull/5013#discussion_r3774733638, this PR adds `oxfmt` support so that it can also be installed directly in a similar fashion without using `npx`.


rami3l:
Also, if one wants to add `oxlint` it should be one search-replace away.

rami3l:
The CI should be green now.

PS: The manifest gen CI error seems unrelated to this particular change.

taiki-e (approved):
Thanks!

taiki-e:
Published in 2.86.0.

rami3l:
@taiki-e Thanks for the quick response and I'm using the new release now!

I have a minor concern though: I can install the tool using your action but the binary name I use has to be `oxfmt-x86_64-unknown-linux-musl` (because that's what they have from the release archives) which might be slightly inconvenient (it's CI-only and we have many ways to work around it afterwards, so no big deal though). Is this already handled by this project somewhere that I unfortunately overlooked, or is this kind of slight optimization something this project wants to support?

Many thanks in advance :)

taiki-e:
Ah, that's a good catch. Filed https://github.com/taiki-e/install-action/pull/1969 to fix it.

taiki-e:
Published the fix in 2.86.1.

